PR #19, opened 2015-01-13
Fix more links

barosl:
I've got curious, what would happen if I make another pull request with the same (but advanced) branch?


Gankro:
:bee:

PR #31, opened 2015-01-28
Miscellaneous changes to TWiR template

iKevinY:
- Link to rust-lang/rust instead of mozilla/rust.
- Italicize *This Week in Rust*.
- Change capitalization of "GitHub".
- Use nouns rather than adjectives to describe the Rust trifecta.

### DIFF
--- a/drafts/YYYY-MM-DD-this-week-in-rust-template.md
+++ b/drafts/YYYY-MM-DD-this-week-in-rust-template.md
@@ -4,13 +4,13 @@ Category: This Week in Rust
 
 Hello and welcome to another issue of *This Week in Rust*!
 [Rust](http://rust-lang.org) is a systems language pursuing the trifecta:
-safe, concurrent, and fast. This is a weekly summary of its progress and
+safety, concurrency, and speed. This is a weekly summary of its progress and
 community. Want something mentioned? [Send me an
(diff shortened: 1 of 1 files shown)


Gankro:
I think the adjectives are more correct (and pleasant).

I read it as 

> pursuing the trifecta (of being): safe, fast, and concurrent

iKevinY:
I think that if going the adjective route, the "of being" should be made explicit.

> …pursing the trifecta of being safe, concurrent, and fast."

Gankro:
Sounds good to me.

brson:
Thanks! I'm always glad for people to help with twir.

The phrase 'pursuing the trifecta: safe, concurrent, fast' is part of an old haiku by @lkuper:

```
a systems language
pursuing the trifecta
safe, concurrent, fast
```

Altering the wording to this extent diminishes the in-joke a bit. That said, I'm happy to take it anyway.

iKevinY:
Ah, I didn't know about the history behind the phrase. I was going the alter the phrasing as per @Gankro's comments but didn't get the chance to since I've been stuck on mobile. If desired, I could open another PR with that change applied, since that version of the phrase is more similar to the original haiku.

lkuper:
FWIW, I don't mind the wording change.  The haiku had a good run. :)

PR #32, opened 2015-01-29
reader-friendly title

indeyets:
Proposing change as [advised by SimonSapin@reddit](https://www.reddit.com/r/rust/comments/2tw2w1/this_week_in_rust_67/co3dmec?context=3)

Original title is instinctively read as "This Week in Rust-67" and I keep wondering what is this "Rust-67" for next 5 seconds
(movie sequels are to be blamed, I believe) :)

### DIFF
--- a/drafts/YYYY-MM-DD-this-week-in-rust-template.md
+++ b/drafts/YYYY-MM-DD-this-week-in-rust-template.md
@@ -1,4 +1,4 @@
-Title: This Week in Rust XX
+Title: This Week in Rust. #XX
 Date: YYYY-MM-DD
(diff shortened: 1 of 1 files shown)


brson:
Thank you for the pull request. I do think I prefer the aesthetics as they are and even like the ambiguity.

I am sympathetic though, and question the utility of sticking a number here at all. Now that we're up to 70 though it would be a shame to stop...

Anyway, the long and short of it is that I am not going to accept this, for vague aesthetic reasons. I apologize, but if you continue to feel strongly about it feel free to re-argue it - I'm easily beaten down with time and perseverance.

PR #35, opened 2015-02-03
Fix typo in username

frewsxcv:
In case you're curious about the misspelling of the car, my name is Corey. I used to like cars...it was clever at the time


Gankro:
Thanks!

PR #43, opened 2015-03-11
Fix typo

fenhl:
prce → pcre


brson:
Thanks for supporting TWiR!

PR #55, opened 2015-04-28
Add Gankro's Peaches post

Manishearth:
Alternatively, if we're okay with linking to poop jokes, http://cglab.ca/~abeinges/blah/everyone-poops/


cmr:
deployed.

PR #66, opened 2015-06-16
[WIP] A rusty makeover.

nasa42:
Design inspired and stolen from blog.rust-lang.org and rust-lang.org.
Currently in draft status. Few issues still need to be resolved like handheld device compatibility, subscription form submission etc. Submitting the PR for early-stage feedback.
This change also introduces one new dependency - https://github.com/miracle2k/webassets (which in turn requires that `sass` is present in `PATH`).

### DIFF
--- a/themes/rusted/templates/article.html
+++ b/themes/rusted/templates/article.html
@@ -1,0 +1,37 @@
+{% extends "base.html" %}
+
+{% block title %}
+{{ article.title }} {%if article.subtitle %} - {{ article.subtitle }} {% endif %} Â· {{ super() }}
+{% endblock title %}
+
+{% block head_description %}
+{% if article.summary %}
+{{ article.summary|striptags }}
+{% endif %}
+{% endblock head_description %}
+
+{% block meta_tags_in_head %}
+{% if article.tags or article.category or article.keywords %}
+<meta name="keywords" content="{{ [article.tags|join(', '), article.category, article.keywords]|join(', ') }}" />
+{% endif %}
+{% endblock meta_tags_in_head %}
+
+{% block content %}
+
+<div class="row text-center">
+  <h3> Like what you see? Subscribe! </h3>
+</div>
+
+{% include "_subscribe-form.html" %}
+
+<div class="post">
+  <header class="post-header">
+    {% include "_post-title.html" %}
+  </header>
+
+  <article class="post-content">
+    {{ article.content }}
+  </article>
+
+</div>
+{% endblock content %}
--- a/themes/rusted/templates/index.html
+++ b/themes/rusted/templates/index.html
@@ -1,0 +1,31 @@
+{% extends "base.html" %}
+
+{% block content %}
+<div class="row">
+  <div class="col-md-12 text-center">
+    <h1 class="pitch">Handpicked Rust updates, <br /> delivered to your inbox.</h1>
+    <p class="subtext">Stay up to date with events, learning resources, and recent developments in Rust community.</p>
+  </div>
+</div>
+{% include "_subscribe-form.html" %}
+<div class="row">
+  <div class="col-md-12">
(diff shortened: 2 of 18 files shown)


nasa42:
Might require some copy editing.

twir-homepage
twir-post
twir-archives

brson:
Design-wise this looks great to me! @cmr what do you think?

themes/rusted/templates/index.html
```diff
@@ -0,0 +1,31 @@
+{% extends "base.html" %}
+
+{% block content %}
+<div class="row">
+  <div class="col-md-12 text-center">
+    <h1 class="pitch">Handpicked Rust updates, <br /> delivered to your inbox.</h1>
+    <p class="subtext">Stay up to date with events, learning resources, and recent developments in Rust community.</p>
```
cmr: "in *the* Rust community"

cmr:
I like the new look too!

cmr:
We'll need a privacy policy before we roll this out.

themes/rusted/templates/index.html
nasa42: It's a bit more complicated than that. If I add four more characters ("the "), it will fall to a second line, so we'll either need to reduce the font or maybe remove "recent ". Or, we can add some more text and make the font larger so that both lines are reasonably wide (and get rid of the `h1` line).
Maybe we can also throw in "... curated by humans ..." somewhere there to add to the length if needed. :-)
/cc @brson

nasa42:
Added a privacy policy.

cmr:
Thanks @nasa42. I'd like to refine it a bit.

nasa42:
@cmr Nope, I do not have any "legal" experience of this sort. If you can get a review from Mozilla's legal department, that would be great!

liquidmetal:
I didn't notice this pull request - I put in a fix for some more responsiveness on the current theme. This new theme looks pretty to me too!

nasa42:
@cmr updated privacy policy changes in https://github.com/cmr/this-week-in-rust/commit/56bf7617b33ab6c202843175b80b7b1a4c49b627

brson:
Where does the privacy policy come from? It seems like *way* overkill for this tiny little website that doesn't collect any sort of data at all.

cmr:
@brson mailchimp requires one to use their service, and the server log does collect data (however miniscule that may be, it counts under EU law (ugh))

nasa42:
@brson as noted in the Git commit, it's a modified version of the privacy-policy found on various Curated (curated.co) operated websites. Here is once such example - http://charged.curated.co/privacy/

cmr:
@brson I think this is good to go if you do.

brson:
Thanks @nasa42! Let's deploy with the next issue.

nasa42:
@brson @cmr please don't use this theme yet - subscription forms don't work atm. I'm still working on a newsletter layout (I'm trying to use [ink](http://zurb.com/ink/), but it uses a `table` based layout and still figuring out how to ask Markdown to emit `tr` and `td`s instead of `p`, `article`, and `section`s.)
I hope to have this theme ready for next week's issue.

cmr:
@nasa42 mind if we just edit out the subscription form? I like the way it looks. Up to you.

nasa42:
@cmr let's just wait till one more issue. :-) Also need to fix a few issues when rendering on mobile devices (last time I checked, button overlapped input field when screen is too small).
Will be ready by next Monday. :-)

themes/rusted/templates/article.html
```diff
@@ -0,0 +1,37 @@
+{% extends "base.html" %}
+
+{% block title %}
+{{ article.title }} {%if article.subtitle %} - {{ article.subtitle }} {% endif %} Â· {{ super() }}
```
mbudde: I think the `Â` should be deleted here.

PR #68, opened 2015-06-20
Added links to posts & projects for next TWiR issue.

nasa42:
* Likely requires copy editing.
* Experimenting with new headings. A few of these can be merged, I think (though keeping them separate for now for feedback). Also, happy to revert to old headings.

### DIFF
--- a/drafts/2015-06-22-this-week-in-rust.md
+++ b/drafts/2015-06-22-this-week-in-rust.md
@@ -15,11 +15,33 @@ If you find any errors or omissions in this week's issue, [please submit a PR](h
 
 This week's edition was edited by: WHO??
 
-# Notable Links
+# From Blogosphere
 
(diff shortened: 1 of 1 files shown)


Gankro:
Oh wow, fantastic! @brson The Dream Is Real

drafts/2015-06-22-this-week-in-rust.md
```diff
@@ -79,7 +101,9 @@ THING is a ...
 
 # Upcoming Events
 
-* [What?]
+* [6/23. Hannover](http://blog.thoughtram.io/rust/2015/06/17/anouncing-hanovers-second-rust-meetup.html)
```
Gankro: I'd like to see these expanded. A location isn't an event.

Gankro:
The new headings seem fun; we're not a rigid formal document.

brson:
This is fantastic! Thank you. You might also update the draft template with these sections.

PR #69, opened 2015-06-20
Add a summary of the changes to rustbyexample over the past few months

mdinger:
@brson Here's a list of the major changes to rbe. I tried to be brief as they can look at the PR if they want more info. I'm not sure if I should have left the dates but it seems okay.

I don't particularly like that the only major changes I found were written by me but that's what I found. There is https://github.com/rust-lang/rust-by-example/pull/583 which is an interesting minor change but it still hasn't merged.


brson:
Thanks!

PR #71, opened 2015-06-23
Adds addition of Herd

imjacobclark:
An experimental HTTP load testing application written in Rust.

Herd is an experimental HTTP load testing application written in Rust. It's main focus is being easy to use and low on OS level dependencies such as the JVM.

Herd is currently able to spawn thousands of concurrent requests to HTTP endpoints by making use of unix process forking and multithreadding.


Gankro:
Thanks!

PR #78, opened 2015-07-07
"All posts" on the home page is now more responsive

liquidmetal:


brson:
Thanks!

I merged this even though I was a bit reluctant to edit the packaged theme (I didn't even realize it was in tree).

PR #87, opened 2015-07-27
Creates ids for headers on articles

bltavares:
With the id attribute on the headers it would be able to link to a
specific section of the page when sharing with someone.

Extension suggestion comes from https://github.com/getpelican/pelican/issues/1357


brson:
Thanks!

PR #91, opened 2015-07-28
add jobs section

oli-obk:
see https://users.rust-lang.org/t/this-week-in-rust-editors-thread/1806/39?u=oli_obk

for now I suggest adding them manually. if it gets more frequent, a thread as suggested by @nasa42 is probably the way to go.

Open question is whether they should get removed in the next or second next issue automatically.

### DIFF
--- a/drafts/2015-07-27-this-week-in-rust.md
+++ b/drafts/2015-07-27-this-week-in-rust.md
@@ -95,6 +95,12 @@ Anderson][brson] for access.
 [erickt]: mailto:erick.tryzelaar@gmail.com
 [brson]: mailto:banderson@mozilla.com
 
+# fn work(on: RustProject) -> Money
+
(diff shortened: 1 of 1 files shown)


brson:
Thanks, I'll merge this into this week's posting and add the section to the template. I like the section name. How long do you think we should let these run?

oli-obk:
For my case 2 weeks, or I'll manually ask not to remove it in the PR that would remove it.

nasa42:
Just like we pull events from calendar, maybe we can pull jobs from [weloverustlang.com](http://weloverustlang.com)? I think @zargony maintains it. Pretty empty though atm, but TWiR can help with that.

oli-obk:
that page looks totally broken on my computer...

zargony:
I'm not the maintainer, sorry. I just discovered that site some months ago..

oli-obk:
uhm... Apparently I mis-translated from german. It should be `student research assistant`. Can I update the currently released version of TWIR?

PR #98, opened 2015-08-03
Fix title of article pages

mbudde:
Remove a `Â` character that has sneaked in.


brson:
Looks like there's a conflict here. Probably because of the temporary style revert.

mbudde:
Made a comment on the pull-request #66 instead.

PR #101, opened 2015-08-10
TWiR release 91 (10 August 2015).

nasa42:
For now, directly added links to some emoticons. Later, should extend Markdown to parse emoticon :tags:.


nasa42:
cc @brson

bluss:
:balloon:  Rust 1.2!! :balloon:

PR #113, opened 2015-09-07
Add the subteam report to 2015-09-07

brson:
r? @nasa42 I'm still not sure the best way to integrate the subteam reports but this seems good enough today.

cc @cmr I can't deploy this today because I am separated from my ssh keys. If you find yourself near yours, and inclined to Rust, will you deploy this edition?


nasa42:
LGTM. :-)

PR #115, opened 2015-09-09
Disable Fira Sans

seankelly:
With Fira Sans as the first font listed, the text is invisible in Firefox on Linux.


nasa42:
@brson this is probably best way to fix current no-text issue. Page doesn't look super-bad either.

brson:
Deployed. Thanks.

PR #117, opened 2015-09-13
Better looking newsletter (specially in Android Gmail).

nasa42:


nasa42:
@brson please merge this, so that we can start sending newsletters from this week. :-)

PR #121, opened 2015-09-15
Send newsletter every Tuesday.

nasa42:
(also fix an issue where a new-line was added to the read-on-web link).


brson:
Sorry for the delay.

PR #123, opened 2015-09-16
fix attribution of QotW

durka:


brson:
Thank you and sorry for the delay.

PR #134, opened 2015-10-06
Fix contributions link in latest newsletter

rgardner:
Even though `drafts/YYYY-MM-DD-this-week-in-rust.md` has the correct contributions link, all recent newsletters seem to use the old wiki link (https://github.com/rust-lang/rust/wiki/Note-guide-for-new-contributors). I only changed the contrib link on the latest newsletter, but I'm happy to update the links on older ones too.


brson:
Thanks!

PR #152, opened 2015-12-14
fix link to RFC 1401

durka:


durka:
This didn't make it to the live site, FYI.

PR #153, opened 2015-12-15
remove Rust-Berlin meetup on 23.12.

hobofan:
The bi-weekly Rust Hack and Learn in Berlin is in holiday mode for the next two meetups (23.12. and 06.01).

After that the Hack and Learns are hosted under http://www.meetup.com/opentechschool-berlin/ instead of http://www.meetup.com/rust-berlin/ .

@skade I think you are the only one who can adjust the events in the [calendar](https://calendar.google.com/calendar/embed?src=apd9vmbc22egenmtu5l6c5jbfc@group.calendar.google.com) since you created them?


skade:
Thanks for notifying, I will change them.

skade:
Changed the dates, also put the meetup in.

PR #159, opened 2016-01-11
Add links to RustBelt

RalfJung:
I am not sure if adding a link in both "News" and "Positions" is appropriate, but it certainly fits both categories. If you'd rather have just one link, I'd vote for "News".


cmr:
I think both is a good idea.

nasa42:
Agree with @cmr, thanks @RalfJung!

PR #166, opened 2016-02-02
Improve Quote Readability

killercup:
If someone really wants a narrow font, they should use a font that is narrow (and not manually adjust the letter spacing for small text sizes).


nasa42:
Thanks!

PR #169, opened 2016-02-09
Create RSS feed and be https

torkve:
I fixed settings and template to create RSS feed for some old readers (e.g. Nokia N9 embedded reader) that know nothing about Atom.

I also fixed SITEURL since the site is fully HTTPS, HSTS-enabled and HTTP-disabled now.


cmr:
Cheers! These'll get deployed with the next issue.

PR #178, opened 2016-03-07
Clarification for drop_in_place

bluss:
Not expected to affect Vec's performance in that aspect (it's already fine).


brson:
Thanks @bluss!

@nasa42 can you redeploy?

PR #179, opened 2016-03-07
Problem: Rust London Meetup #4 not in schedule

vladimir-lu:
Solution: Add it


brson:
@nasa42 Can you deploy?

PR #188, opened 2016-03-21
Eddyb is not a new contributor

Manishearth:
(he's been around for ages)


nasa42:
We use a script to fetch all new contributors, so he probably made a commit with his real name (or different spelling).

PR #190, opened 2016-03-22
Add 3/22 SF Bay Area meetup

erickt:


nasa42:
Probably late for this but thanks! :-)

PR #192, opened 2016-03-28
easy -> less easy

jdm:
The issue's always been marked `E-less easy`, it was just incorrectly categorized in TWiR.


nasa42:
Thanks!

PR #193, opened 2016-04-04
TWIR 125 CotW and notable changes

llogiq:
I only counted changes on rust-lang/rust this time. Is that OK?


nasa42:
Yes, it's ok I guess.

PR #195, opened 2016-04-11
a minor documentation change is not notable

TimNN:
rust-lang/rust#32774 is a minor documentation change and not major implementation change (which is the obvious interpretation of the current phrasing (at least in my opinion).

By the way - feel free to just close if you disagree and / or don't want to change a published issue.


llogiq:
One time I don't look at the diff and not one but two people send PRs. :smile:

:+1:

nasa42:
Thanks!

PR #196, opened 2016-04-11
Fix mistakes in the latest TWIR

petrochenkov:
Private-in-public is not a lint, it's only *sometimes* a lint for backward compatibility because the previous checker implementation was buggy, it will turn into an error soon.
Calling it a lint gives a wrong impression that it may be `allow`ed.

Unions are defined with a contextual keyword `union`.


nasa42:
Thanks!

PR #199, opened 2016-04-13
Add Rust Belt Rust CFP reminder announcements

carols10cents:
Hi! I'd love to have Rust Belt Rust's CFP in TWiR until April 30, when it's closing-- I've put it under News but if you think it would be a better fit under Events then I'd be happy to move it (or have it moved for me, whichever :)) Let me know!! Thanks!!!


nasa42:
I'll keep it in news section for next 2 weeks. Thanks!

PR #205, opened 2016-04-26
Fixed grammar, added Darmstadt meetup

llogiq:


nasa42:
Nope, this is not the first _These Weeks in Rust_. :-)
Here is one from past - https://this-week-in-rust.org/blog/2013/11/09/these-weeks-in-rust-22/

PR #215, opened 2016-05-10
UnsafeCell::get_mut instead of `as_mut`

oli-obk:
[libs team decided on `get_mut`](https://github.com/rust-lang/rust/pull/32565#issuecomment-217041987)


nasa42:
Thanks!

PR #221, opened 2016-05-23
twir-131's CotW and notable changes

llogiq:
Same procedure as every week. Cheers! :smile:


nasa42:
Wrong target file, but I've fixed that manually now. :-)

PR #222, opened 2016-05-23
Fix code markdown formatting.

frewsxcv:


nasa42:
Wrong target file, but I've manually fixed that now. Thanks!

PR #242, opened 2016-06-28
Replace :smile: with 😴

llogiq:
jekyll does not expand emojis, so...


nasa42:
We use Pelican not Jekyll. http://getpelican.com/ :wink:

PR #244, opened 2016-07-01
TWiR next

brson:
Help wanted bugs, a new job listing, and a new fott.

I've been sad that the friend of the tree tradition has died. Putting them into the subteam reports hasn't worked out very well because they don't get much visibility. So I'd like to try putting them directly into twir. Let me know if you think this is inappropriate.


brson:
cc @nasa42 @llogiq

brson:
I'm going to think about this a bit.

nasa42:
@brson This looks good to me. May I merge?

PR #246, opened 2016-07-04
Added S.U.P.E.R. Android Analyzer

Razican:
We just published an Android applications analyzer written in Rust. It's extensible, really fast and detects lots of vulnerabilities and bad practices. It's still in an alpha stage, but it's already better than usual tools in many aspects.


nasa42:
Thanks!

PR #248, opened 2016-07-06
fix generics &lt; + &gt; problem

fbstj:
The `ThinAttributes` change to `ThinVec<Attributes>` had &lt; and &gt; that aren't correctly displayed


nasa42:
Thanks!

PR #252, opened 2016-07-12
I didn't edit this week's TWiR

llogiq:
Next week, I'll return.


nasa42:
I've made some changes so that it is obvious that you are a regular editor. :-)

PR #261, opened 2016-07-27
add a PhD position to the job offers

oli-obk:
as noted in the comments, only show it once, in the next twir, do not roll it over to the next week.


nasa42:
Thanks!

PR #266, opened 2016-08-02
Add note that imag has a website, irc channel and mailinglist

matthiasbeyer:


matthiasbeyer:
I hope this is appropriate and I'm not to late to the party :smile:

PR #271, opened 2016-08-07
Updates for imag

matthiasbeyer:


matthiasbeyer:
I'm not sure whether I should add https://github.com/matthiasbeyer/imag/issues/245 as issue ... I consider it hard to do, even if one is familiar with the imag codebase...

nasa42:
I suggest to add with a `[hard]` label.

matthiasbeyer:
Okay, I will update this PR for it.

PR #277, opened 2016-08-19
Replace mangled angle brackets with curly braces

emosenkis:


nasa42:
I'm not sure of this format either. I'll copy the heading from https://github.com/rust-lang/rust/pull/35447 verbatim.

nasa42:
Thanks for the PR!

PR #281, opened 2016-08-23
Remove issue that is already WIP

matthiasbeyer:
That was faster than expected.

This issue is already WIP, so we can remove it here.


nasa42:
Nice!

PR #290, opened 2016-09-02
Fix layout releasetag+bytestool

frehberg:


frehberg:
Please merge this patch, it is fixing enumeration of entries bytestool and releasetag

PR #297, opened 2016-09-07
Fix denver meetup date

brson:


brson:
cc @llogiq @nasa42 Can someone redeploy?

brson:
The event was scheduled incorrectly on the calendar, which I've fixed.

llogiq:
I cannot because I don't have any rights on the server.

nasa42:
Deployed.

PR #306, opened 2016-09-16
upgrade releasetag 0.6.0

frehberg:


nasa42:
Sorry, I could not include releasetag in this week's issue as it is already mentioned in last week's issue.

PR #347, opened 2016-11-15
Add a short paragraph that novemb.rs is happening

skade:


nasa42:
I'm moving this section below `Blog Posts` because IMHO, readers who are accustomed to the format of TWiR posts likely skip anything above `Blog Posts` as persistent blurb.

PR #358, opened 2016-12-18
Pegasus and WebMetal announcements, plus gfx-0.13

kvark:


nasa42:
Hi @kvark, after some thought, I've decided to remove Pegasus and gfx-rs announcement lines as we no longer publish New Crates section and release 0.13 doesn't seem to be a major change.

Thanks for the PR!

kvark:
@nasa42 I wonder why you stopped doing the "New Crates"?

gfx-0.13 may not be the biggest release in history, but we've got quite a few features shipped (transform feedback, persistent mapping, proper errors, and Metal backend), so it's far from a maintenance release either. I don't mind the policy of limiting the project updates in general, but it seems quite unfair to me that various This Week In XXX posts are getting into TWIR without questions, while the scale of changes is incomparable to gfx-0.13, given that it's been developed for 5 months since gfx-0.12.

nasa42:
It's a bit of a blurred line between what qualifies for inclusion and what doesn't. I'm sorry that it looked unfair to you that your crate news was not included.
The other weekly news articles are added mostly because of convention as they have been a continuous feature of TWiR for more than a year.
I really don't have a specific set of objective criteria on what to include and what not to, and it mostly falls down to my own subjective opinion. We should come up with a _notability guidelines_ for TWiR, but unfortunately I don't have a complete answer on how to define that.

kvark:
@nasa42 👍 for notability guidelines
Since it would be fairly difficult to formalize them, we could at least make them simple. For instance - no crate updates. At the end of the day, it's crate author's call to whether announce an update or not. And if they consider the update to be notable - they need to crate a blog post of sorts, which automatically includes all the TWiX.

PR #366, opened 2017-01-02
Adding link to crates.io RFC-ish PR

carols10cents:
Hi! I have [this crates.io PR](https://github.com/rust-lang/crates.io/pull/488) that's not really an RFC but I'd like input from the community on it. I'm trying to wrap it up, so I'm declaring a final comment period and I'd love to have it in this week's TWiR.

It looks like this file still has last week's FCP RFCs in this section, is it ok if I just add mine at the bottom here? Will it get overwritten when this gets updated to this week's?

Feel free to move the link to another section if it isn't appropriate here since it's not a real RFC!! Thanks!!! ❤️


nasa42:
I'll move it to the top of FCP list for more visibility.

PR #372, opened 2017-01-12
Forensic-Tool Development with Rust

getreu:
Abstract: Within the framework of this study the suitability of the Rust ecosystem for forensic tool development was evaluated. As case study, the tool Strings­ext was developed. Starting from analysing the specific requirements of forensic software in general and those of the present case study, all stages of the software development life-cycle have been executed, up to the first production release. Strings­ext is a reimplementation and enhancement of the GNU-strings tool, a widely used program in forensic investigations. Strings­ext recognizes Cyrillic, CJKV characters and other scripts in all supported multi-byte-encodings while GNU-strings fails in finding these in UTF-16 and other encodings.

    During the case study it has become apparent that the Rust ecosystem provides good support for secure coding principles and unit testing. Furthermore, the benchmarks showed a satisfactory performance of the resulting Strings­ext binaries comparable to the original C version.

### DIFF
--- a/drafts/2017-01-17-this-week-in-rust.md
+++ b/drafts/2017-01-17-this-week-in-rust.md
@@ -16,6 +16,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ## News & Blog Posts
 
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
This is awesome! I hope I can use it as source in my own masters thesis! I also hope I can write rust code for my masters thesis (and hopefully as well in a job after my graduation)! :smile:

xen0n:
Also using Rust in startup project & master's thesis here 😄 (although it's only used to accelerate things for Python, but not the very topic; still it's awesome to have Rust featured in thesis!)

PR #373, opened 2017-01-16
Add 'answer of the week'

Timmmm:
I've had two amazing answers from people on SO where the quality is such that I wish it were in the official documentation. Anyway I think it would be nice to highlight this useful information, and maybe encourage more people to answer Rust questions.

Alternatively it could just go in the News & Blog Posts bit like last time.

### DIFF
--- a/drafts/2017-01-17-this-week-in-rust.md
+++ b/drafts/2017-01-17-this-week-in-rust.md
@@ -26,6 +26,10 @@ This week's Crate of the Week is [trust](https://github.com/japaric/trust), a Tr
 
 [submit_crate]: https://users.rust-lang.org/t/crate-of-the-week/2704
 
+# Answer of the Week
+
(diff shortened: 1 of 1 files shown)


nasa42:
While this sounds interesting, it's quite an effort to scout Stack Overflow each week to come up with an _Answer of the Week_. Would you be willing to commit to finding an _Answer of the Week_ every Monday?

There is also the side-effect of promoting Stack Overflow as an outlet for Rust community. Which isn't a bad thing really but it also raises the question of why not also have a _Discussion of the Week_ to promote official Rust forums?

I'm personally not in favour of a new section for this. That doesn't however mean that TWiR is not interested in highlighting outstanding community contributions happening on Stack Overflow. I suggest that it goes in _News & Blog Posts_ section.

Timmmm:
Ok fair enough, news/blog posts it is!

PR #387, opened 2017-02-03
fotf nominations for 2017-02-07

erickt:
Congrats @Yamakaky!


erickt:
And congrats to @colin-kiegel and @Florob!

PR #405, opened 2017-03-08
TWIR 173: Add Reddit Turing Challenge

mattico:


mattico:
Could perhaps change the wording to clarify that the type system _is_ turing complete.

PR #409, opened 2017-03-22
174: remove Cornell job listing

durka:
It doesn't mention Rust at all.


durka:
Maybe they changed the description? Or the link is wrong?

nasa42:
I've changed it back to the Twitter link and will remove in next issue.

PR #411, opened 2017-03-26
style RFCs

nrc:


nasa42:
I'll move "Good first issues" to "Call for Participation" section. I think they will get more visibility there.

nasa42:
Keeping "Good first issues" where they are for now as "Call for Participation" is already quite crowded this week.

PR #431, opened 2017-04-27
Add PX8 Rust project

hallucino:


nasa42:
Hi @hallucino, it's probably a better idea to directly link to issues on which you are expecting participants, rather than a link to the repository.

hallucino:
Yes you're right !

It was most for the 'News & Blog Posts', is it possible ?

nasa42:
We try to avoid announcing new crates in "News & Blog Posts" section. I'd suggest you to nominate it for [Crate of the Week](https://users.rust-lang.org/t/crate-of-the-week/2704) instead.

hallucino:
Thanks !

PR #447, opened 2017-05-24
No CotW, but at least some cool PRs

llogiq:
Sorry for being late.


nasa42:
This seems like a late night PR! Thanks!

PR #448, opened 2017-05-24
Add the link to TiKV weekly update

queenypingcap:
TiKV weekly update is missing.


nasa42:
I skipped these because these weeklies have been mostly about minor updates and also because [there was some discussion](https://github.com/cmr/this-week-in-rust/pull/358#issuecomment-268541912) about skipping announcements of seemingly non-minor releases for some crates while still including these weeklies containing minor updates.

To be honest, I still don't have an objective criteria on what to include and what to skip, and deciding on that is one of the most difficult part of my TWiR routine.

For now, I've decided to only link to other weeklies published by core team. Sorry about that.

PR #450, opened 2017-05-30
CotW and notable changes

llogiq:
Less changes, but more CotW this week...


nasa42:
Unfortunately, [oath2](https://crates.io/crates/oath2) is yanked from crates.io so I removed it from CotW.

PR #466, opened 2017-07-04
Add RustFest Zürich

dns2utf8:
The CFP is open too: https://cfp.rustfest.eu/


nasa42:
I think it's probably better to put the CFP announcements in News or Call for Participation section for better exposure.

dns2utf8:
Ah, good point. Will change it. Thank you.

PR #481, opened 2017-07-26
fix rustfest cfp deadline

durka:


durka:
Also cc @skade because the linked page says both August 1 and July 31 in different places.

nasa42:
The date is correct but the link was wrong. I've now fixed that.

PR #483, opened 2017-07-26
replace emoji images with unicode

durka:
I dunno if everyone's browser supports these, but the hotlinked images from Discourse [won't be loaded anyway](https://www.reddit.com/r/rust/comments/6pkc8b/this_week_in_rust_192/dkqdx83/).


nasa42:
Thanks for the PR!
I'm unsure about browser compatibility (in Ubuntu, it didn't look good and will probably need another font to be included), so I'll stick to Discourse CDN emojis for now.
We're working on resolving the CSP issues and images rendered properly.

PR #484, opened 2017-07-31
Add the Herald

skade:
We never announced the Herald here (and we also had a bug to fix), so here's an announcement :).


nasa42:
Awesome!

PR #493, opened 2017-08-09
add my blog post

durka:
I wasn't sure how the list was sorted so I just put it in the middle next to another post from the same day.


durka:
Either is fine by me.

durka:
#495

PR #497, opened 2017-08-15
Add dev-tools news to twir

nrc:
Is it OK to add this much text? I've cut down quite a bit of detail from twidt. I think in the future we could just have a link, but for the first issue or two it would be nice to have some text here to get us some publicity.


nasa42:
Ok if I just link to the blog post?

nasa42:
I added this as a link in issue 195.

PR #504, opened 2017-08-23
fix typo of petrochenkov in 2017-08-22 Update

KodrAus:
Hi there!

I _think_ there's a typo in one of the _who's awesome_ peeps and it should be for @petrochenkov?


KodrAus:
cc: @nasa42

PR #509, opened 2017-08-30
fix spelling of Abomonation

durka:
It's really abomonable.


matthiasbeyer:
The old wasn't correct? From what I see (via dict.cc) the word "abomonation" does not even exist!

durka:
It's not a word but it is the name of the crate :)

On Wed, Aug 30, 2017 at 1:40 PM, Vikrant Chaudhary <notifications@github.com

PR #516, opened 2017-09-10
added 'Implementing the Read trait for strings'

Leopard2A5:


nasa42:
I think this is already possible with [`BufReader`](https://doc.rust-lang.org/std/io/struct.BufReader.html)? Also see - https://github.com/rust-lang/rust/issues/10564

Leopard2A5:
Hmm, the way I see it `BufReader` only wraps a `Read`, and since String doesn't implement `Read`...
But I see now that `Read` *is* implemented for `&'a [u8]`, so if i just convert the String to that I can put it into a `BufReader`. Thanks for the links!

PR #556, opened 2017-11-28
No CoTW, at least notable changes

llogiq:


nasa42:
Maybe we should make `faster` CotW? (it was [nominated few weeks back](https://users.rust-lang.org/t/crate-of-the-week/2704/316)).

PR #559, opened 2017-11-29
Add meetup in Amsterdam

moretea:
We have a meetup about Procedural Macro's and Custom Derive :)


nasa42:
You might also want to add the event to today's issue.

moretea:
@nasa42 done!

PR #560, opened 2017-12-03
Added a link to "Zemeroth's devlog 2"

ozkriff:


nasa42:
Usually we avoid release notes and weeklies (with some exceptions), but I think we can make the title more interesting to include this one. :wink:

PR #565, opened 2017-12-09
Add Extending C with Rust and Guide to Diesel ORM

spejss:


nasa42:
Thanks for the PR!
I looked at the [rust/c article](http://spejss.com/index.php/2017/12/03/extending-c-with-rust-101/) but I found it to be quite simplistic and something that is already covered by [Rust FFI Omnibus](http://jakegoulding.com/rust-ffi-omnibus/) and [Rust FFI examples](https://github.com/alexcrichton/rust-ffi-examples).
I've removed the rust/c post, but will be including Diesel ORM post in this week's issue! (I changed the title a bit).

PR #594, opened 2018-01-28
Add new post

ayrat555:


nasa42:
Hi, @ayrat555. An interesting read. However, unfortunately, it may not be a good fit for TWiR. To keep our focus on technical articles and news, we try to avoid certain articles including ones those that are along the lines of _Why I love/hate Rust_. We had a lot of these in past, and it is unlikely that any new information can be extracted from new articles on these topics.

I'd suggest to also post your article on [Rust Herald](herald.community.rs/) and [r/rust subreddit](https://www.reddit.com/r/rust/)!

Thanks for the PR!

PR #597, opened 2018-02-01
Add link to "Trying Rust" video

mjgpy3:


nasa42:
Hi, @mjgpy3. An interesting video. However, unfortunately, it may not be a good fit for TWiR. To keep our focus on technical articles and news, we try to avoid certain articles including those that are along the lines of Why I love/hate Rust or getting started with Rust. We had a lot of these in past, and it is unlikely that any new information can be extracted from new articles on these topics.

I'd suggest to also post your video on Rust Herald and r/rust subreddit!

Thanks for the PR!

mjgpy3:
Hi @nasa42, that makes perfect sense, thanks for looking at this.

PR #599, opened 2018-02-07
add blog post: Sharing for a Lifetime

RalfJung:
Seems this one got missed. :)

I have no idea how these are sorted, so I just added it somewhere.


nasa42:
Thanks for the PR! Could you please add a short sentence after the link describing what the post is about? "Sharing for a Lifetime" is clever but not very descriptive. :-)

RalfJung:
Fair enough. ;) How about this?

PR #638, opened 2018-05-07
Added "Rust: Enterprise Services Need Not Suck"

paulirotta:


nasa42:
Hi, @paulirotta. An interesting article. However, unfortunately, it may not be a good fit for TWiR. To keep our focus on technical articles and news, we try to avoid certain articles including those that are along the lines of Why I love/hate Rust or getting started with Rust. We had a lot of these in past, and it is unlikely that any new information can be extracted from new articles on these topics.

Thanks for the PR!

(related [reddit thread](https://www.reddit.com/r/rust/comments/8hq7sb/rust_enterprise_services_need_not_suck/)).

PR #640, opened 2018-05-14
Move Berlin's Birthday-Hack'n'Learn a day early

badboy:
We moved this week's hack'n'learn to a day earlier.


badboy:
Force-pushed with updated meetup URL

nasa42:
I've merged it but it's probably already too late.

badboy:
Thanks anyway! There are still 3 hours for people to join us!

PR #655, opened 2018-06-12
adding contribution to twir, Actix::From(Rocket)

noyez:


nasa42:
I've made the title a bit less-fancy for better readability. :slightly_smiling_face:

PR #656, opened 2018-06-12
CotW, QotW & notable changes

llogiq:


nasa42:
Thanks for also taking care of QotW!

PR #658, opened 2018-06-18
Add link to lpc82x-hal 0.2 announcement

hannobraun:


nasa42:
We usually don't do release announcements, so I've tweaked the title a bit.

hannobraun:
Thank you!

PR #661, opened 2018-06-19
CotW, notable changes, a meetup and QotW

llogiq:


nasa42:
QotW while funny, I wonder if it would be considered negative towards C++. :man_shrugging:

llogiq:
I see it as pretty tongue-in-cheek. *Someone* has to complain after all...

PR #672, opened 2018-07-11
Add "Why Rust is Important" Blog

alexcameron89:
I wrote this blog with the intent on helping non-Rust users understand the purpose and benefits of Rust. It's purposely high level and stays away from code so that it is approachable for a larger audience.

### DIFF
--- a/drafts/2018-07-17-this-week-in-rust.md
+++ b/drafts/2018-07-17-this-week-in-rust.md
@@ -16,6 +16,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ## News & Blog Posts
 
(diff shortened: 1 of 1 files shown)


nasa42:
An interesting article. However, unfortunately, it may not be a good fit for TWiR. To keep our focus on technical articles and news, we try to avoid certain articles including those that are along the lines of - why I love/hate Rust or getting started with Rust. We had a lot of these in past, and it is unlikely that any new information can be extracted from new articles on these topics.

Thanks for the PR!

PS: You may also want to consider posting this article on [r/rust subreddit](https://www.reddit.com/r/rust).

PR #681, opened 2018-07-26
Fix date in filename. Add Lehi, UT meetup.

CleanCut:
Fix date in filename. Add Lehi, UT meetup.


ErichDonGubler:
You da man, thank you!

PR #696, opened 2018-08-22
Changed all api.github.com links to github.com

Darsstar:
See https://www.reddit.com/r/rust/comments/999jbu/this_week_in_rust_248/e4m2ryr/


qezz:
It looks like in [drafts/2018-08-28-this-week-in-rust.md](https://github.com/cmr/this-week-in-rust/blob/master/drafts/2018-08-28-this-week-in-rust.md) some of the links under the "Final Comment Period" are broken, too. 

I think we should file an issue.

PR #705, opened 2018-08-29
Add a missing blog post

RalfJung:
Is there a list of feeds or so you are checking? In case you're not aware, [ReadRust](https://readrust.net/all/) does an excellent job of aggregating Rust blog posts.

Given that the content to send a PR against usually only appears very shortly before publication, there is only a tiny window that I have to hit if I want to ask to be included in TWiR.


nasa42:
You can always find the next week's issue in "drafts/" directory, so it's usually 7 days of window to create a PR. :wink:

RalfJung:
Well last few times I checked, the blog post section of the draft was always empty. Good to know you are accepting PRs for it anyway :)

PR #743, opened 2018-11-02
add news on Martin release

annabarinova:


nasa42:
Thanks for the PR!

To direct TWiR's focus on technical articles (and important community news), we avoid including opinions pieces and crate announcements (with some notability exceptions). As such, unfortunately, we may not be able to accept this PR.

I'd however encourage you to post this article on [Rust subreddit](https://www.reddit.com/r/rust/) and also to [Crate of the Week](https://users.rust-lang.org/t/crate-of-the-week/2704) thread which is followed by people interested in new crate announcements!

PR #775, opened 2018-12-20
Readd some jobs

oli-obk:


oli-obk:
Just one more week is fine, thx

PR #776, opened 2018-12-23
Adding rust interview link

agbell:
Adding link to interview here  https://corecursive.com/024-software-as-a-reflection-of-values-with-bryan-cantrill/


nasa42:
Thanks for linking to the video! However it seems like interview is not about Rust, and Rust is only a small part of the conversation.

Additionally, to direct TWiR's focus on technical articles (and important community news), we avoid including opinions pieces and release announcements (with some notable exceptions). As such, this PR may not be a good fit for TWiR.

I'd however encourage you to post this article on [Rust subreddit](https://www.reddit.com/r/rust/).

agbell:
Makes sense, thanks.

PR #798, opened 2019-01-18
Add link for LeetCode

Brooooooklyn:
LeetCode now support Rust 🎉
Add link for leetcode.com and leetcode-cn.com


nasa42:
Could you please point to an authoritative source that claims that Rust support is a recent addition? Otherwise it's hard to verify for reader. E.g., an official blog post or changelog that mentions recent Rust support.

PR #806, opened 2019-02-05
Add 18. Feb Karlsruhe meetup

Lireer:


nasa42:
I think we should link directly to the event page, and I couldn't find it at https://www.meetup.com/Rust-Hack-Learn-Karlsruhe/
I think we should wait until they create the event page, so we can link to it directly.

Lireer:
I have added a link directly to the event page, should I open a new PR or is it possible to add to this one?

PR #834, opened 2019-02-26
added link "Run Rust code in Pandoc Markdown documents" under News

gpoore:
I've created a program that makes Rust code in Markdown documents executable, and I think this might be of interest to the community. This is sort of like R Markdown for Rust, if you're familiar with that. This can be useful for writing things like tutorials. Here's an [example of what is possible](http://htmlpreview.github.io/?https://github.com/gpoore/codebraid/blob/master/examples/rust.html).

### DIFF
--- a/drafts/2019-02-26-this-week-in-rust.md
+++ b/drafts/2019-02-26-this-week-in-rust.md
@@ -16,6 +16,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ## News & Blog Posts
 
+* [Run Rust code in Pandoc Markdown documents](https://github.com/gpoore/codebraid)
+
 # Crate of the Week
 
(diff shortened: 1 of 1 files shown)


nasa42:
Thanks for the PR!

To direct TWiR's focus on technical articles (and important community news), we avoid including opinions pieces and release announcements (with some notable exceptions). As such, this PR may not be a good fit for TWiR.

I'd encourage you to post this article on [Rust subreddit](https://www.reddit.com/r/rust/) and also to [Crate of the Week](https://users.rust-lang.org/t/crate-of-the-week/2704) thread which is followed by people interested in new crate announcements!

gpoore:
@nasa42 Thanks for the suggestions! I'll look into posting in those locations. I had misunderstood what TWiR is looking for in submissions.

PR #858, opened 2019-03-12
Usual Sections

Flavsditz:
Maybe there is a way of doing this in a better way: when I see the dates in the calendar they are usually relative to my time, so depending on where the meeting is it might show it for me a day later or earlier.

That also happens of course if I access the calendar using the API: it will get on GMT time. Do you know a way of accessing the events based on the time zone they were created?


nasa42:
I usually visit the link and find the date on the page. Yup it's more manual process, and I wonder if we can streamline it further.

PR #879, opened 2019-04-02
Add async article

gruberb:
Reddit: https://www.reddit.com/r/rust/comments/b8hi0s/explained_how_does_async_work_in_rust/


nathany:
Thanks Bastian. I've been wanting to learn more about Rust async/await.

Btw, your Go example is demonstrating a goroutine (green threads multiplexed on OS threads) but isn't an example of asynchronous I/O. The Go runtime hides the details of async I/O so it's kind've of hard to show -- opening a file or database connection just looks like ordinary synchronous code.

gruberb:
Hm true, thank you @nathany! I will update the text so it's more clear!

PR #895, opened 2019-04-16
C/QotW+notable changes

llogiq:


nasa42:
> Confusion is a product type.

I'm not so sure about the QotW in today's issue. Without complete context, the quote is quite... confusing. I'm thinking of removing QotW from today's issue, but I'd be happy to revert if you feel otherwise.

llogiq:
In that case, either add the context or perhaps I may suggest [this tweet](https://mobile.twitter.com/XMPPwocky/status/1117162034779623424)?

nasa42:
I'm probably not smart enough to understand either quote, so I'd let you decide. I'm happy to merge any PR you make. :slightly_smiling_face:

llogiq:
To explain it (&let you in on the joke): peano is a crate that implements numbers as types by counting. This is very inefficient, and so the type system was easily overwhelmed even on simple operations (e.g. 8×4).

The response is funny because this is a type which represents a product, so a 'product type' (which means a struct in Rust lingo, as opposed to an enum which is a 'sum type').

PR #904, opened 2019-05-01
Fixed cast-rs link

therealprof:
Though I wonder whether we should use https://crates.io/crates/cast instead...


nasa42:
Thanks!

PR #923, opened 2019-05-26
Adding three blog posts by Sean Policarpio (kdrakon)

kdrakon:


nasa42:
I'd suggest to post these on [r/rust subreddit](https://www.reddit.com/r/rust) for scrutiny. And if received positively, I'd be happy to include it!

PR #942, opened 2019-06-24
Request for implementation

jonathan-s:
There is the github repo request for implementation, where you can submit issues of a crate or broader issue that you would like for someone to solve. 

This can can be a grow-ground for new rustaceans. Perhaps it would be a good idea to make this repo a recurring feature of the newsletter?


nasa42:
I'm not sure if it makes sense to add link in each single issue (there is already too much boilerplate in issues), but I'd be happy to mention in once in a while!

PR #953, opened 2019-07-10
move period outside of link to fix broken url

programmerjake:


programmerjake:
@cmr note that this should be fixed before this week's email is sent out.

PR #963, opened 2019-07-19
Added video of whiteboard series on Rust Libp2p

nearmax:


nasa42:
Thanks for the PR!

The video seems to be more about Libp2p and not Rust in general.
Additionally, we usually prioritise text articles, as such I think [Rust Videos](https://twitter.com/RustVideos) might be a better fit to include this video.

I'd also suggest to post it on [r/rust subreddit](https://www.reddit.com/r/rust) to reach a wider audience. :+1:

PR #971, opened 2019-07-30
Add Rust in blockchain call for participation

brson:
cc @Aimeedeer


brson:
Ping @llogiq @nasa42 why didn't this make it in? Should I submit again for next week?

PR #987, opened 2019-08-20
Adding blog post "First Amazon Alexa Skill in Rust"

muditc:


nasa42:
The blog post seems to only introduce Amazon Alexa and not talk about "how" of Rust integration, as such it may not be a good fit for TWiR. (Additionally this topic is already covered in TWiR through https://medium.com/@amalec/building-alexa-skills-in-rust-4cf54a497ea4).

Thanks for the PR!

PR #991, opened 2019-08-29
Add links to next/previous article on article page

### LINKED ISSUE #986: Ability to browse next and previous issues
It would be great to have links to the next and previous (if applicable) issues on the article page. I just discovered this newsletter (which is great by the way!), but there is currently no way for me to continue browsing once I go to an article page without going back to the archive page.

I can submit a PR if that would be helpful. :)

Thank you for spending your time on this!

RussianCow:
Resolves #986. Here's an example screenshot:

image

The CSS uses flexbox for layout, but it should degrade gracefully in IE < 11 like so (untested):

image

### DIFF
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -15,7 +15,7 @@ else:
     THEME = 'themes/rusted'
 
 PLUGIN_PATHS = ["plugins"]
-PLUGINS = ['assets']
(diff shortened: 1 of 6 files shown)


nasa42:
This is amazing! Thank you!

PR #1021, opened 2019-10-01
C/QotW+notable changes

llogiq:


programmerjake:
I was under the impression that algebraics would be the crate-of-the-week since it has 6 votes to pin-project's 3 votes.

see also https://github.com/cmr/this-week-in-rust/pull/1014#issuecomment-535231476

llogiq:
Oops, must have overlooked it. I'll change the PR.

llogiq:
Ah, wait a minute: algebraics was suggested the week before, and lost out to texture-synthesis. That's why I didn't include it. We usually don't look into last week's suggestions. Feel free to re-suggest, though.

nasa42:
I've renominated algebraics for next week. :+1:

programmerjake:
> I've renominated algebraics for next week.

Ah, ok. Thanks!

Issue #1017, opened 2019-09-26
Responsive theme for mobile phones


dvogt23:
Is there a way to make the newsletter content more responsive for mobile phones? I have always to zoom and swipe left/right to read.

nasa42:
This seems to work for me on Gmail app. Which email app do you use and could you please paste a screenshot?

dvogt23:
I use the Mail on my iPhone. Cause of the small I have too zoom...
![B9D999FB-DE0B-4C29-8002-F3C48E6E0AD2](https://user-images.githubusercontent.com/9352647/65684625-59d31a00-e060-11e9-9e9c-be30ecacd979.png)

dvogt23:
Another cool idea would be the new darkmode style for the newsletter theme?

dvogt23:
#1035 I dont know why, but on my iPhone, the font color is grey not white 😑 hows look like in other mail clients?

nasa42:
I use Gmail on Android. Unfortunately I can't test it because [dark theme for Gmail is supported in Android 10](https://gsuiteupdates.googleblog.com/2019/09/dark-theme-android-ios.html), which I at present don't have access to.

nellshamrell:
Closing due to the age of this issue, but if it is still a problem please feel free to open a new issue.

PR #1035, opened 2019-10-10
Add dark mode

dvogt23:
Issue: #1017


dvogt23:
Here a screenshot:

nasa42:
I am not sure if this will work though as some major email clients only use inline CSS. Let's test in today's newsletter.

PR #1039, opened 2019-10-14
add new blog post about script programming

Aimeedeer:


nasa42:
Thanks for the PR!

I read through the article and it seems to be an article mainly about CKB VM and not Rust programming language (or its ecosystem). As such, I'm unsure if it would be a good fit for TWiR and I'd close the PR for now, please let me know if you think otherwise.

PR #1043, opened 2019-10-17
Update 2019-10-22-this-week-in-rust.md

crd477:
Add reference to blog post


nasa42:
An interesting article. However, unfortunately, it may not be a good fit for TWiR. To keep our focus on technical articles and news, we try to avoid certain articles including those that are along the lines of - why I love/hate Rust or getting started with Rust. We had a lot of these in past, and it is unlikely that any new information can be extracted from new articles on these topics.

Thanks for the PR!

PR #1074, opened 2019-11-18
Add @scy's upcoming Advent of Code livestream

badboy:
@scy is planning to do coding livestreams to solve Advent of Code puzzles, in German.
Maybe this should be pushed into next week though to be closer to when it starts?

### DIFF
--- a/drafts/2019-11-19-this-week-in-rust.md
+++ b/drafts/2019-11-19-this-week-in-rust.md
@@ -16,6 +16,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ## News & Blog Posts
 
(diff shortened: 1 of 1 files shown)


scy:
Thanks for submitting this! 😊 Well, I am planning to do some streams in November too: setting up Rust on my machine and going through [Rust by Example](https://doc.rust-lang.org/rust-by-example/) or some other tutorial.

But publishing it in TWiR next week sounds better nevertheless. Even if I did a few streams by then, it’s probably better to have a few videos to catch up to than nothing to watch at all. (And there certainly won’t be anything to watch until tomorrow when TWiR comes out.)

nasa42:
I think the announcement fits better in "Events" section, I've moved it there.

PR #1089, opened 2019-11-28
WIP Theme support

flying-sheep:
Needs work to support links and code, but first I wanted to know if the approach is OK:

Since this is a dynamic property, I use CSS variables to avoid having to duplicate all rules using a variable inside of a @media block. I renamed the dynamic colors to a more semantic gradient of 5 greys and a brand color.

Honestly we can probably leave the brand color static, or we reuse it for links. Also the primary button doesn’t seem to use the brand color at all anyway?

I also added some code to make it work with Python 3 (which lacks cssmin) and newer Pelican versions.

grafik

### DIFF
--- a/plugins/assets/assets.py
+++ b/plugins/assets/assets.py
@@ -32,7 +32,10 @@ except ImportError:
 def add_jinja2_ext(pelican):
     """Add Webassets to Jinja2 extensions in Pelican settings."""
 
-    pelican.settings['JINJA_EXTENSIONS'].append(AssetsExtension)
+    if 'JINJA_ENVIRONMENT' in pelican.settings:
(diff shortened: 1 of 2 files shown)


nasa42:
Thanks! I'm happy to accept this. There is already a dark-mode for newsletter theme. Probably best to use same colours here?

nasa42:
Link to dark-mode PR: https://github.com/cmr/this-week-in-rust/pull/1035.

flying-sheep:
Ah, sorry for not replying: I don’t really like the complete white-on-black there. It works for small OLED screens, but even then, I’m not a fan of that white headerbar.

Do we have any other precedent? Any dark theme done by a designer instead of me or @dvogt23?

nellshamrell:
Going to go ahead and close this due to age of the issue. Feel free to open this again if you wish to update it :)

PR #1095, opened 2019-12-10
C/QotW+notable changes

llogiq:


nasa42:
I feel like replacing QotW as it can be considered slight bit demeaning towards C++?
What do you think?

PR #1099, opened 2019-12-12
fix link to rust-lang/rust#66606

tanriol:


llogiq:
I really need to add some regex to check against this to my workflow.

PR #1100, opened 2019-12-12
Scale the newsletter appropriately on mobile email clients

Aloso:
Scales the newsletter appropriately and consistently on mobile email clients.

Tested in Chrome Web Tools:

mobile-newsletter

This relies on only 1 CSS rule in an `@media` block, so it should look fine in email clients that don't support `@media`.

### DIFF
--- a/themes/newsletter/templates/base.html
+++ b/themes/newsletter/templates/base.html
@@ -1,14 +1,15 @@
 <!doctype html>
-<html>
+<html lang="en">
   <head>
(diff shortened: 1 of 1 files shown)


Aloso:
I'm sorry, I accidentally committed this with the wrong message.

nasa42:
Tested with Mailchimp/Litmus and seems to be looking good! Today's emails will have these CSS changes. Thanks!

PR #1122, opened 2020-01-14
Propose QotW

matthiasbeyer:
From the file history, there's no QotW just yet, so I'll propose one here.


nasa42:
Quotes go through a voting process before they are selected as QotW.
Please nominate here: https://users.rust-lang.org/t/twir-quote-of-the-week/328

matthiasbeyer:
Ah, dang, now I remember. Sorry for the noise.

PR #1126, opened 2020-01-22
Add Rust LATAM conference call for paper reminder

apiraino:
Hope it's ok to put this in the "News & Blog Posts" section, I'm linking our Reddit thread.

thanks!


nasa42:
I've moved the link to "Call for Participation" section as historically we've linked Call for Papers in that section and it might get more exposure there (I'll put it at top of the list). Let me know if you think it would be more apt to move it to News section.

apiraino:
@nasa42 good idea, thanks. I was unsure where to put it.

PR #1137, opened 2020-02-06
Added an article on interactive applications

Michael-F-Bryan:


spacejam:
words like "sanity" might distract from your core message

Michael-F-Bryan:
@spacejam, how so?

spacejam:
more and more in recent times, words like "sane" and "insane" are being interpreted as words that reinforce the notion that disabled people are inferior. personally, I used to use those words frequently, and I was surprised when people started mentioning to me that I might be offending people. but I see that those words have been used to discriminate against people, and that by avoiding those connotations, it's easier for my actual intended message to be heard by people who have experienced discrimination of this sort. 

then I realized that a pretty sizable proportion of the best engineers I know are likely people who have faced this kind of discrimination, and that's when it clicked for me that I would be able to reach the people I want to reach more easily by avoiding those terms.

Michael-F-Bryan:
I get where you're coming from although I don't necessarily agree with the interpretation.

It took me a while to articulate why, but I think the word "sane" (as in, ["in a state of mind which prevents normal perception, behaviour, or social interaction"](https://www.lexico.com/en/definition/insane)) and "maintain your sanity" (["
to stop someone worrying or becoming mentally ill"](https://www.macmillandictionary.com/dictionary/british/keep-someone-sane)) have different meanings. Just because the title contains the word "sane" doesn't mean it has anything to do with labelling people or implying superiority/inferiority. It may just be our different life experiences, but I've actually found people to be pretty understanding of mental well-being and people who may have it tougher than others.

Normally if I were talking to a coworker I'd say something like "creating interactive applications which won't make future Michael sad" because (from personal experience) poorly thought out code and technical debt can be quite demoralising when you need to work on them in the long term (the "worrying" and "become mentally ill" bit). But that doesn't really roll off the tongue... Can you think of a more appropriate title for the article?

Michael-F-Bryan:
I don't want to double-dip, so I'm closing this PR in favour of #1142.

PR #1140, opened 2020-02-10
Adding blog post

gterzian:
Adding https://medium.com/@polyglot_factotum/further-thoughts-on-async-await-7767f924cb7e

Thanks!

### DIFF
--- a/drafts/2010-02-11-this-week-in-rust.md
+++ b/drafts/2010-02-11-this-week-in-rust.md
@@ -16,6 +16,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ## News & Blog Posts
 
+* [Further thoughts on async/await.](https://medium.com/@polyglot_factotum/further-thoughts-on-async-await-7767f924cb7e).
+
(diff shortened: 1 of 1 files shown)


nasa42:
Thanks for the PR!

The article seems to have received negative votes on [reddit](https://www.reddit.com/r/rust/comments/ezq09a/further_thoughts_on_asyncawait/) with concerns raised on misinformation, and as such it may not be a notable article for wider Rust community.

I'd recommend to review suggestions in reddit comments to help refine the article. Closing for now.

gterzian:
That's disapointing. 

So you don't want to include articles that "reddit" disagrees with? I actually never look at comments on reddit, because they are usually total nonsense. I will not change this good habbit now. 

Do you have any guidelines on how to avoid censorship? Anywhere I can appeal? 

You've published a ton of articles by me before, so it would be weird if this article somehow represented "misinformation". 

If I disagree with most people, that doesn't make it "misinformation". Unless you believe in  the dictatorship of the majority, or something like that.

PR #1182, opened 2020-04-03
Add paris

0x20F:
Add link for paris


nasa42:
Thanks for PR!

To keep our focus on technical articles and news, we try to avoid mentioning new crates in News & Blog Posts sections.

We do have a different outlet "Crate of the Week" to promote crates. I'd suggest to nominate your crate for CotW at https://users.rust-lang.org/t/crate-of-the-week/2704.

PR #1187, opened 2020-04-08
Add blog post

bishtpawan:


nasa42:
Thanks for the PR!

I read the post and I've found it to be a bit lacking in technical details (e.g., post makes a claim that `await` can only be used inside `async` block but doesn't explain why, doesn't mention why async-std crate is being used etc.). Additionally the grammar can also be improved.

I'd suggest to improve the article, get it peer-reviewed, and submit again.

PR #1188, opened 2020-04-09
Remove explaine.rs as it is duplicated

jrvidal:


nasa42:
I'm about to publish the issue and have already removed the article.
Thanks for the PR!

PR #1189, opened 2020-04-12
new blog post entries for 334

hmble:
I got some of this links from [Read Rust](readrust.net) and some from [subreddit](reddit.com/r/rust)


nasa42:
Thank you for collecting these! I might remove link to "Swift: Google's bet on differentiable programming", as it doesn't directly relates to Rust.

PR #1202, opened 2020-04-23
Add Rust and AWS Lambda blog post

jamesmcm:
This is a blog post I wrote covering my first experience deploying Rust to AWS Lambda - converted to an example so the readers can follow along. 

The non-trivial parts covered are:

* Using a custom deserialiser for some of the fields
* Converting Excel dates (from Calamine) to chrono::NaiveDate
* Getting SSL to work with Redshift
* Cross-compiling with OpenSSL on OS X


nellshamrell:
Looks great!

PR #1203, opened 2020-04-23
Add ALMA PR to Calls For Participation

jamesmcm:
[ALMA](https://github.com/r-darwish/alma) is a Rust application to create persistent Arch Linux Live USB installations.

There are many open issues, this is what I considered to be the most important (mainly to raise awareness in the Rust developer community, as many readers will be Arch Linux users).

Note I am a significant contributor to ALMA, but not the project founder.


nellshamrell:
Awesome!

PR #1210, opened 2020-04-27
Add blog post

koivunej:
Link: https://medium.com/equilibriumco/rust-ipfs-append-only-log-001-42106983f3c


nellshamrell:
Hello there! We generally do not include new crate or new app version announcements as there are so many of them they could easily overrun the newsletter. This feels very IPFS specific - there's not a lot of general content that could apply to a majority of TWiR readers. It feels like there is a small amount of content about testing and working on a distributed team - but again it's very IPFS specific. To make it more broadly applicable, I would love to see more info about how you developed the integration tests with more code snippets of how you implemented them - beyond the changes you made to keep the endpoints the same. Content about integration testing - especially deep dives into how it can be set and used with lots of code examples - is something that is needed in the Rust Community. I'm going to close this pull request for now - but I would welcome submissions of future articles that are a bit more general and broadly applicable.

koivunej:
@nellshamrell thanks a lot for the ideas and the feedback! Agreed, such content would be needed.

PR #1217, opened 2020-04-30
Fix the link to Crust of Rust by Jon Gjengset

alexander-jackson:
The link in the recent issue points to a spreadsheet instead of the actual
YouTube video. It also should be annotated with the [video] tag like the link
above it.


nellshamrell:
Thank you for doing this! It's already been fixed through another commit, but I really appreciate this!

PR #1229, opened 2020-05-06
Finesse the Rust/WinRT info

chris-morgan:
The spelling and capitalisation were wonky, and the ™ symbol out of place.
This makes things match reality and Microsoft's descriptive conventions.


nellshamrell:
Thank you!

PR #1240, opened 2020-05-11
[Blog] Post on collecting multiple errors from an interator of results

tarquin-the-brave:
Hi,

I wrote this blog post looking at some ways of collecting multiple errors from an iterator of Results. It's a problem I've solved in a few different ways over the past couple of years when it's cropped up. The post does include a solution using a crate that I made a couple of weeks ago to solve this problem, discussing the pros and cons against other approaches.  

I see from past PRs that you don't tend to accept posts that a straight up announcements of new crates.  If you feel that's what this is, no worries.  I thought this might be useful to some people and would be interested if anyone has feedback.  I couldn't find an approach like this from searching around.

https://tarquin-the-brave.github.io/blog/posts/collecting-all-the-errors/

Thanks

### DIFF
--- a/drafts/2020-05-12-this-week-in-rust.md
+++ b/drafts/2020-05-12-this-week-in-rust.md
@@ -16,6 +16,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ## News & Blog Posts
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hi @tarquin-the-brave! Thank you for submitting this - I feel there is enough content exploring error collection in Rust beyond the crate announcement to include it in the newsletter. Merging!

PR #1252, opened 2020-05-18
Add Rust releases for single and multiple targets with GitHub Actions…

mrcosta:
I wrote a tutorial about how to add Rust releases for single and multiple targets with GitHub Actions. I hope you like it.


nellshamrell:
This is great, thank you!

PR #1275, opened 2020-05-31
Added "Rust Bindings for LittlevGL"

lupyuen:
Thanks :-)


nellshamrell:
Hello! While this is certainly cool (and I commend you for making it!) we don't generally include links to crate announcements (whether GitHub repos or others). How would you feel about writing up a blog post about how this library does the awesome things it does with Rust? What new insights about Rust did you gain when you were writing it? I would LOVE to feature an article like that.

rafaelcaricio:
@nellshamrell Sounds good. This is definitely something I should write about. Thank you! Some motivation to write a new blog post. :)

PR #1309, opened 2020-06-16
Add Rust Search Extension v0.10

Folyd:


nellshamrell:
We don't generally include crate announcements in this newsletter (because there are so many of them). I would love to see a blog post about HOW this tool uses Rust - what did you learn about Rust in the process of creating/updating it - and how it enhances the experience of Rust users.

PR #1312, opened 2020-06-17
Add "Introducing GameLisp" to the "News & Blog Posts" list

fleabitdev:
Thanks 👍

### DIFF
--- a/draft/2020-06-23-this-week-in-rust.md
+++ b/draft/2020-06-23-this-week-in-rust.md
@@ -18,6 +18,8 @@ Check out [this week's *This Week in Rust Podcast*]()
 
 ## News & Blog Posts
 
+* [Introducing GameLisp, a scripting language for Rust game development](https://www.reddit.com/r/rust/comments/h13657/)
+
 # Crate of the Week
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
This looks very cool! However, we don't generally include announcements of new crates (because there are so many of them!). Would you be willing to write up a blog post of some kind 1) Introducing glsp and 2) Focusing on HOW glsp does what it does - how does it build on Rust to allow you to do more?

fleabitdev:
It's a fair suggestion, but I don't yet have a blog, or any real intention to start blogging about GameLisp.

It occurs to me that, within GameLisp's reference manual, the [Introduction for Rust Programmers](https://gamelisp.rs/reference/introduction-for-rust-programmers.html) explores your suggested topic: GameLisp's relationship to Rust, and how it builds on Rust to enable the programmer to do more. It's not a blog post per se, but I'd have no objections to a list item like:

- [GameLisp: Introduction for Rust Programmers](https://gamelisp.rs/reference/introduction-for-rust-programmers.html)

If that won't be possible, then no worries.

PR #1322, opened 2020-06-23
Add bincode blog post

ZoeyR:


nellshamrell:
Hello, thank you for the submission! We don't generally include crate/library updates because there are so many of them. However, if you tweet this and tag the ThisWeekInRust twitter account, I would be happy to retweet it!

PR #1332, opened 2020-06-29
Invite to Academic Survey

jacobkrueger:
Hey,
We are conducting an academic survey as stated in the News section (which is hopefully the right one). We have been told by the R Core Team that this would be a suitable option to contact R developers who may be interested in participating (instead of spamming developers by mail).
Best,
Jacob


nellshamrell:
Hello there! As this isn't Rust specific, I'm hesitant to include it in the newsletter. However, if you tweet this link and tag @thisweekinrust I will be sure to retweet it!

PR #1344, opened 2020-07-06
Add post about kernel modules in Rust

thedataking:
Link to our latest blog post on transpiling kernel modules to Rust.


nellshamrell:
Looks like this was already merged in #1345 - thank you for the contribution!

PR #1370, opened 2020-07-19
Add reference to MRML

jdrouet:


nellshamrell:
Hello there! We don't generally include links to GitHub repos in the newsletter. However, if you were to write up a blog post on HOW you made this re-implementation - particularly information about how you used Rust and what you learned while doing it - I would love to feature it!

PR #1380, opened 2020-07-22
Update 2020-07-21-this-week-in-rust.md

crd477:
Fix small closing tag


not-matthias:
Just noticed that too and wanted to make a PR, but you were faster. Good catch. :+1:

nellshamrell:
Thank you!

PR #1384, opened 2020-07-23
Remove benchmark reference

richardpringle:
Remove benchmark reference with questionable results and without code to prove the validity of the benchmark.


richardpringle:
I think the results of that benchmark are misleading. The code on which the benchmark is not provided to validate the results either.

This benchmark finds very different results and code is provided. If it is valid, it confirms my suspicions about the original benchmark that was posted.
https://www.reddit.com/r/grpc/comments/hv7h6i/grpc_bench_a_clear_objective_grpc_benchmark/

nellshamrell:
Hello there - I'm going to leave this in, but if you would like to put together a blog post with your own benchmarks I would be happy to include it in a future issue of TWiR.

PR #349, opened 2016-11-28
No CotW, but some notable changes

llogiq:
Aand I'm back. :smile:

PR #1397, opened 2020-07-28
TWiR #349

cdmistman:
I did a few experimental things...

1. I added subsections in the "Updates from Rust Community" section
2. To make that section cleaner (and because "News and Blog Posts" was the only 2nd-level section in there), I got rid of the "News and Blog Posts" subsection and made all subsections direct subsections of the section (sorry for word vomit, might be easier to see in the file)
3. Changed non-english language links to use 2-letter ISO code instead of the English name for the language.

If anything should be changed let me know!


nellshamrell:
Hi there! Thank you so much! I love #1 and #3 - but I'm wondering if we might keep the News & Blog post heading and make the labels subsections under News & Blog Posts. I tend to prefer this to make it more distinct from the following sections (CFPs, RFCs, etc.)

cdmistman:
What if we change the subsections for "Updates from Rust Community" to be 3rd-level? This would bring the style closer to the current "Upcoming Events" section. I'm ok with keeping the "News & Blog Posts", I'm just wary of the redundancy since that's the only subsection of "Updates from Rust Community"

nellshamrell:
Ah....I see what you mean. That - changing the subsections of "Updates from the Rust Community" to be 3rd level - sounds good to me!

PR #1411, opened 2020-07-31
Added writeup on SIMD acceleration for nucleotide encoding/decoding

Daniel-Liu-c0deb0t:


nellshamrell:
Hello! This certainly looks like a great project! When I'm considering project write ups like this one, I'm looking for more information about how they relate specifically to the Rust language - what did you learn about Rust when writing the project? What can others learn about Rust from your project? There are SO many fantastic Rust projects out there, unfortunately there are many more than we can include unless there is information specific to the Rust language. Would you be willing to include more Rust specific (what you learned from Rust, what features of Rust you included, etc.) information in the write up? I'm going to close this PR for now, but feel free to open it up again if you add content like that to this post :)

PR #1427, opened 2020-08-05
twir 350

cdmistman:
- Since there were almost 30 posts in the learning section, I broke it up into "Learn Standard Rust" (ie vanilla, the Rust you learn without introducing frameworks or you're experimenting with the lang) and "Learn More Rust" (with crates, non-local target compiles, complex topics, etc)
- Moved Rusty Days video to Misc since it's not _just_ tutorials and stuff (the video would generally fall under "learn more rust" but Steve's and Tim's talks are more under "observations/thoughts")
- Fixed some formatting


nellshamrell:
These are great! Thank you!

PR #1460, opened 2020-08-20
event(online): add the rust munich meetup event on 25th of august

sassman:


sassman:
unfortunately this event is already over. 

What is the good timeframe to submit this sort of PR in time to get merged before the issue of TWiR is issued?

PR #1507, opened 2020-09-13
Add rust vs java testing link

mrcosta:
I've written a blog post about my personal experience and perception of the testing approach between Java and Rust. Let me know what you think


nellshamrell (changes requested):
Great post, but I think it would fit better under the Observations/Thoughts section. Could you move the link to be under that heading?

mrcosta:
Done =)
Let me know if it's ok now.

nellshamrell (approved):
Looks great! Ty!

PR #1512, opened 2020-09-16
TWiR 356

cdmistman:
In the second commit, I introduce a new top-level section just below News & Blog Posts for highlighting the 2021 roadmap CFP. We can discuss further how to do this.


nellshamrell:
Most excellent as always! Thank you!

PR #1522, opened 2020-09-22
Add "Porting EBU-R128 audio loudness analysis from C to Rust"

sdroege:
Not sure if that's best under "Observations/Thoughts" or further down under "Misc". This is the initial blog post of a 4-part series about porting some C code to Rust. This part is the general overview, the next parts would fit best under "Learn More Rust".

Feel free to move it where you think it fits better :)


nellshamrell (approved):
Thank you!

PR #1530, opened 2020-09-24
Update links.txt: TWiR 357, 2020-09-23

hmble:
Also removed duplicate entry from TWiR 356


nellshamrell (approved):
Great work as always, thank you!

PR #1533, opened 2020-09-25
Added blog posts

bishtpawan:
1. OS in Rust: An executable that runs on bare metal: Part-1
2. OS in Rust: An executable that runs on bare metal: Part-2


nellshamrell (approved):
Thank you!

PR #1548, opened 2020-10-05
Use statement thoughts

tarquin-the-brave:
Hi,

A colleague ask me why don't put all my `use` statements at the top of the page and the resulting discussion went on way longer than either of us expected.  So I thought it would be a good thing to put together a blog post on.


nellshamrell (approved):
Thank you!

PR #1557, opened 2020-10-08
Remove duplicated link

pickfire:


HoLLy-HaCKeR:
This should be a link to [here](https://blog.rust-lang.org/inside-rust/2020/10/07/1.47.0-prerelease-2.html) instead

PR #1593, opened 2020-10-26
Add RustFest Global schedule

skade:
Added the RustFest Global schedule announcement. I didn't quite find the right spot to put it, so I thought "well, it is kind of an out of the order announcement, so let's have a section".

Feel free to rearrange.


nellshamrell (approved):
Thank you!

PR #1594, opened 2020-10-27
Increase the font weight

### LINKED ISSUE #708: Very light font can be difficult to read
Thanks for this awesome resource!  I have a suggestion to make:

I'd like to recommend against using such a light font for all body items on "TWIR".  It makes things very difficult to read sometimes. 

image

I could be wrong, but this looks like Helvetica or Arial with a very light weight.  I would suggest at least bumping up the weight to 400, or even changing the font to something more legible in general (like Lato, [InterUI](https://rsms.me/inter/), etc.)

equt:
This PR should fix #708.

- `body` element is now set to `normal`
- all headings are set to **`bold`**

### DIFF
--- a/themes/rusted/static/css/_base.scss
+++ b/themes/rusted/static/css/_base.scss
@@ -17,7 +17,7 @@ body {
     font-family: $base-font-family;
     font-size: $base-font-size;
     line-height: $base-line-height;
-    font-weight: 300;
+    font-weight: normal;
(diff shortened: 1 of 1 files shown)


nellshamrell:
Thank you!!!!

zofrex:
Thank you! That's a lot easier to read on my screen, thank you so much for taking the time to improve it.

PR #1600, opened 2020-10-28
Update 2020-10-28-this-week-in-rust.md

simon-void:
I wanted to learn how to use Rust (wasm modules) from JavaScript, but all the examples I could find either used webpack or where incomplete. Since I wanted to understand how this works without the added complexity of webpack (or any other bundler), it took a bit longer for me to find a solution. But the solution I found is as easy as it gets I think, only 1 html file, 1 js file, 1 rust file and 1 toml file is enough! So the link points to the GitHub Page I created which points to its own source and is hopefully a valuable resource for people how are looking for a Hello-World level, proof of concept of using Rust code in the browser.

### DIFF
--- a/draft/2020-10-28-this-week-in-rust.md
+++ b/draft/2020-10-28-this-week-in-rust.md
@@ -23,6 +23,7 @@ No newsletters this week.
 ### Observations/Thoughts
 
 ### Learn Simple Rust
+- [Minimal setup of JS code invoking Rust/Wasm code](https://simon-void.github.io/minimal_js_rust_setup/)
 
 ### Learn More Rust
 


nellshamrell:
Hello! Thank you for this contribution. I definitely see what you are trying to do here - show a demonstration of what people can do and the source code that will enable them to do that. I'm wondering if you might be willing to write this up a little more as a blog post, including insights which you learned about Rust and WASM along the way? The solution is great, but I think our readers will interested in your journey to get there.

simon-void:
thanks for your feedback. I can see where you are coming from. The problem is that I don't have a blog, but I do have some experience with Hugo website generator and GitHub Pages, so I might just be able to set one up. but it'll take some time. Let me get back to you in a week or two.

nellshamrell:
Sounds good! Thank you! Closing this for now, but feel free to re-open in the future :)

PR #1602, opened 2020-10-31
Add link to Semantic FFI Bindings in Rust - Reactivating the Borrow Checker

Tamschi:
Hopefully I'm doing this right.

I recently put Rust on my wristwatch (custom target, i.e. easier on Nightly) and ended up writing a bit about (mostly) extern types.
I think the blog post would be interesting to the wider Rust community, in that I haven't seen anyone previously write about the patterns this feature enables in practice (beyond also theoretically correct raw pointer types).


nellshamrell (approved):
Looks great! Ty!

PR #1615, opened 2020-11-05
No longer sends every Tuesday

Urcra:
Seems like it's been quite some time since I last received the newsletter every Tuesday, so let's update the information about that


nellshamrell:
Fair point, and ty!

PR #1616, opened 2020-11-05
Update Repo URL

dns2utf8:
So people are less confused where they can submit their news.

Cheers,
Stefan


nellshamrell (approved):
Thank you!

PR #1621, opened 2020-11-08
Add Truelayer job postings

apiraino:
Adding Truelayer 2 job postings

(@nellshamrell wasnt sure if job postings are accepted *only* or *also* via twitter, sorry if this PR is inappropriate)


nellshamrell (approved):
This is fine :) Thank you!

PR #1634, opened 2020-11-16
Add blog post introducing xd crate

delan:
Apologies if the symbols aren’t allowed! I thought they would be nice to have as a hint for what “code page 437” means, but it’s not the end of the world.

* https://www.azabani.com/2020/11/15/xd.html
* https://crates.io/crates/xd


nellshamrell (approved):
Thank you!

PR #1643, opened 2020-11-19
Add post discussing ways of generating config file references

tarquin-the-brave:
Hi,

I did some experimenting with different ways of generating reference files for tools' config.  Here's my findings.

Thanks!


nellshamrell (approved):
This is great! Ty!

PR #1644, opened 2020-11-19
Adding a link to blog post for fwumious (and not github)

andraztori:
@nellshamrell I hope this works then :) A post in the same format as others... 

Somehow I thought you would prefer raw github link... don't know why.


nellshamrell (approved):
Looks good! Ty!

PR #1657, opened 2020-11-27
Propose a new quote of the week…

bwinton:


llogiq:
We take proposals on the [rust-users thread](https://users.rust-lang.org/t/twir-quote-of-the-week/328) so people can vote on them. Please submit your suggestion there.

bwinton:
Cool, good idea!

PR #1667, opened 2020-12-02
TWiR 367

cdmistman:
Also removed a Rust Insiders post that got carried over from last week


nellshamrell (approved):
Fantastic as always! Ty!

PR #1672, opened 2020-12-07
include Dec 10 Rust DC event

scooter-dangle:
Had put it in the community calendar, but it was missed somehow. (Maybe due to it originally having been scheduled for November?)


nellshamrell:
Can do!

PR #1680, opened 2020-12-10
move "Interior Mutability" post to walkthroughs section

ibraheemdev:
it looks like my post that I added in #1676 was accidentally moved into "project updates" my a later commit. This pr moves it back to "walkthroughs"


nellshamrell (approved):
Sorry about that, ty for the correction!

PR #1696, opened 2020-12-17
add article "Learning Rust: static trait bounds"

ericseppanen:
A discussion of what it means when we need `'static` on a generic type, even when no references are present in the code.

This is something that seemed a bit mysterious to me at first, and I couldn't find a satisfying explanation-- so I thought I'd try to write one.


nellshamrell (approved):
Looks great! Ty!

PR #1697, opened 2020-12-18
Add Allsorts year in review post

wezm:
We (YesLogic) published an overview of what's been happening with [Allsorts](https://crates.io/crates/allsorts) over the last year.


nellshamrell (approved):
Thank you!

PR #1700, opened 2020-12-18
Add "How to write a Terminal Multiplexer with Rust, Async, and Actors"

austinjones:
Add a post about [tab](https://crates.io/crates/tab), and how async can power very low-latency apps.


nellshamrell (approved):
This is great! I love the info about the Actor model.

austinjones:
Yay!  Thanks for the merge :)

PR #1720, opened 2021-01-04
Add blog post about nonymous + bore(1)

delan:
I wasn’t sure whether to put this in Project/Tooling or Observations/Thoughts (or Miscellaneous). It’s technically a new/updated crate post, but the vast majority of the content is a deep dive into designing no-alloc-friendly DNS codecs.


nellshamrell (approved):
Thank you!

PR #1744, opened 2021-01-16
Add link to serial_int crates.io page

JKAnderson409:
This is a tiny crate I was working on this week for serial integers. It's very simple but I think it can be useful for a lot of people.


nellshamrell (changes requested):
Hi there! It certainly looks very cool, but we don't generally include links to crates or GitHub repos without more context. Would you be willing to write up a blog post explaining what the crate adds to the experience of a Rust user? And maybe something about what you learned about Rust from writing it? That info is so useful to our readers!

JKAnderson409:
OK, I’ll do a write-up explaining its purpose and usefulness and change the PR to a link to that when it’s ready.

nellshamrell (approved):
This is perfect! Ty!

PR #1750, opened 2021-01-20
added xaynet

Robert-Steiner:
thank you


nellshamrell (changes requested):
Hi there! It certainly looks very cool, but we don't generally include links to crates or GitHub repos without more context. Would you be willing to write up a blog post explaining what the crate adds to the experience of a Rust user? And maybe something about what you learned about Rust from writing it? That info is so useful to our readers!

Robert-Steiner:
Hi @nellshamrell 
I'm sorry I didn't know. A blog post is a good idea👍
Thanks a lot!

PR #1757, opened 2021-01-24
Automatic flamegraph generation with Criterion

Jibbow:
Hi there!
Thanks for running TWiR! It's the first time I contribute to this project. I hope that's the right process and I hope that adding this blog post is alright :)


nellshamrell (approved):
Thank you!

PR #1761, opened 2021-01-26
Add a Rust job in Vermont

seamusabshere:
Faraday is an equal opportunity employer and we value diversity at our company.

If you think this is cool, don't rule yourself out because you're missing something from this list - we want people with imposter syndrome to apply.

Don't have a resume ready? (presumably because you're happy at your job and weren't considering anything) Please just fill out https://forms.gle/g76sDDrDb5eYU5jL6


nellshamrell (approved):
Thank you!

PR #1787, opened 2021-02-03
C/QotW + notable changes

llogiq:
Also added a job offer at aleph alpha.


nellshamrell (approved):
Thank you! Great work as always!

PR #1789, opened 2021-02-03
Add new Job posting for AMD Driver Engineer

Chris--B:
We are continuing to grow our team and looking to hire another
developer with Rust experience to work on tooling for AMD GPUs. 😊

This is the same team that was advertised on the 2020-09-30 Issue.


nellshamrell (approved):
Thank you!

PR #1800, opened 2021-02-10
feat: include Feb Rust DC meetup link

scooter-dangle:
Hey! I was late in adding the event to the calendar, so I thought I'd submit an explicit PR in case you've already run your calendar-scraping script.


nellshamrell (approved):
Looks great! Ty!

nellshamrell:
I'll let you in on a secret...I currently copy all of the events from the calendar into the newsletter by hand (sometimes the dates move around for the meetups or the meetup skips a month, I always check the meetup page to verify that it is running). Thanks for saving me some keystrokes!

PR #1803, opened 2021-02-12
Fix disposition for option_unwrap_none

sersorrel:
It looks like Option::expect_none(msg) and unwrap_none() are in fact not being stabilised: https://github.com/rust-lang/rust/issues/62633#issuecomment-777009512


nellshamrell:
Thank you!

PR #1810, opened 2021-02-16
Add 'Writing our own CE' to misc

Lonami:
Hi! I've taken on a journey to complete [Cheat Engine's](https://cheatengine.org/) Tutorial program with Rust. I'm aiming for a new entry to the series every weekend. So far, two entries are out, and I've added a link to the second one, because the title is slightly more catchy and it encourages people to read the prior entry.

I think it would be a good addition because it dives into the topics of FFI, unsafe code, constructs like `MaybeUninit`, and overall is a good challenge.


nellshamrell (approved):
Looks great! Ty!

PR #1864, opened 2021-03-11
Add an event that missed the publication deadline

naufraghi:
We missed the deadline for 381 :cry:  but wed better stay with other events in the same date range.

If not possible, I'll update the PR to be included in the next draft.

Thanks!


nellshamrell (approved):
No problem, I'll add this in.

nellshamrell:
This was added separately

PR #1869, opened 2021-03-13
Add config-maint

matthiasbeyer:
I'm not 100% positive whether this belongs in the newsletter or not, but I figured a PR wouldn't hurt :smile: Feel free to reject if you don't think that this should be in the newsletter.


matthiasbeyer:
Please don't include this, this information is outdated. Config is maintained again.

PR #1872, opened 2021-03-16
2021-03-17: Add "How to send emails with Rust" to Walkthroughs

skerkour:
Hello, I've added my short tutorial about how to send emails in Rust the the Walkthroughs section :)


nellshamrell (approved):
Thank you!

PR #1875, opened 2021-03-16
Add "rkyv is faster than {...}" blog post

djkoloski:
The title is a reference to [burntsushi's famous ripgrep post](https://blog.burntsushi.net/ripgrep/), hopefully it's not considered offensive.

This blog post is mostly about a serialization benchmark suite I made and the results from it. Many of the serialization library maintainers (prost, bincode, postcard) have been excited to have a thorough benchmark to use for performance testing.


nellshamrell (approved):
This is fine - I went through it and there were no personal attacks and nothing I considered a rant. Thank you!

PR #1884, opened 2021-03-18
Fixed the link for totally_safe_transmute.

insideoutclub:


nellshamrell:
Merged in #1882, which also included a fix, but thank you so much for contributing!

PR #1888, opened 2021-03-19
Add 'Introduction to Apache Arrow with Rust'

andrewleverette:


nellshamrell:
It looks like the images for the code examples are broken - I'm not able to see them in either Firefox or Chrome.

andrewleverette:
Hey @nellshamrell, I tested on Chrome and Safari both mobile and desktop versions. That link displays the images correctly for me. I also tested with JavaScript disabled just in case that might be an issue, but I got the same results. I'm not sure why you wouldn't be able to see them.

andrewleverette:
Hey @nellshamrell. Could you elaborate on what you're experiencing? I've been trying to duplicate what you have described, but every way I try the images are loading correctly.

nellshamrell:
Well...now it's working. I blame it on something weird on my end. Will merge!

PR #1893, opened 2021-03-23
2021-03-24: Add 'How to execute shellcodes from memory in Rust' to Rust Walkthroughs

skerkour:
Hello, I've added a tutorial about how to execute shellcodes from memory in Rust to the Walkthroughs section :)


nellshamrell (approved):
Thank you!

PR #1919, opened 2021-04-06
2021-04-07: Add "How to create small Docker images for Rust" to Walkthroughs

skerkour:
Hello, I've added a post about "How to create small Docker images for Rust" in the "Walkthroughs"

Cheers ✌️


nellshamrell (approved):
This is great! Ty!

PR #1925, opened 2021-04-07
Add Microsoft Job to 2021-04-07 Edition

rylev:
Let me know if adding the job as a PR is ok or if you'd rather collect it through tweet. Unfortunately the job description only mentions it as a senior role, but there is also room for juniors on the team.


mastermaxy:
also wondering the same as @rylev, don't want to be bothersome with PRs :)

nellshamrell:
Submitting them in PRs is welcome!

nellshamrell (approved):

nellshamrell:
Thank you!

PR #1927, opened 2021-04-07
Add try_trait_v2 RFC, which just hit FCP

scottmcm:
Figured I'd sneak this in so TWiR-watchers will get more notice about it.


nellshamrell (approved):
Thank you!

PR #1962, opened 2021-04-23
Add a link to blog post "Rust notebooks with Evcxr"

deontologician:
This is about half observations / thoughts, and half an interview with the creator of Evcxr.
Not sure what the best title to use is, feel free to modify it


nellshamrell (commented):
Looks great! Ty!

PR #1985, opened 2021-05-03
Add "Compilers as teachers" blog post

skade:
A small blog post on how a diagnostics issue was found in a training and how it was solved before the training was over :).


nellshamrell (approved):
Ty!

PR #1996, opened 2021-05-05
fix dead link

efx:
I did not change it in the `links-upto352.txt` because my editor would
introduce other unrelated changes.

The prior URL, notes.willcrichton.net, did not resolve to a valid IP. Something DNS.


nellshamrell (approved):
Thank you!

PR #2001, opened 2021-05-06
Remove duplicate entry from 389

kvark:
The other entry is called "Making Generative Art with Rust: interview with Alexis André"


murlakatamenka:
Late Night Confessions appears twice too.

nellshamrell (approved):
Thank you!

PR #2006, opened 2021-05-07
Added Blog on Rust Embedded 

NitinSaxenait:


nellshamrell:
Thanks for submitting this, but it is very similar to an article that is already in the draft https://blog.knoldus.com/why-rust-for-embedded-development/

Since the article in this pull request does not have a lot of content on Rust (it's more of an intro to embedded systems), I'm going to just go with the previous one. Closing this, but thank you again!

PR #2013, opened 2021-05-11
Added a blog on STM32 Discovery Board

NitinSaxenait:


nellshamrell:
This is not Rust specific, so unfortunately I need to close this PR. I appreciate your enthusiasm for sharing your posts in This Week in Rust, but I'm wondering if you might be willing to limit it to 2 articles per issue? I want to make sure we feature a lot of different publications and authors.

PR #2014, opened 2021-05-11
Rust DC May 2021 meetup + recording of April 2021 meetup

scooter-dangle:
:wave:

Regarding the recording, is `Miscellaneous` the correct section for that?


nellshamrell (approved):
Looks great! Ty!

PR #2026, opened 2021-05-14
Fix: Article heading says 700, not 7000

matthiasbeyer:
Article already published, not sure what your policy is on updating published stuff...


nellshamrell (approved):
We welcome updates! Thank you!

PR #2035, opened 2021-05-19
Added Behind the scenes of 1Password for Linux [using Rust]

beyera:
I think @steveklabnik [said it best](https://twitter.com/steveklabnik/status/1394714711321628678) about [this post](https://dteare.medium.com/behind-the-scenes-of-1password-for-linux-d59b19143a23) (no paywall).
> very happy to read this post; it's just as much "how #rustlang was adopted at 1password" as it is "behind the scenes of 1password on Linux"


nellshamrell (approved):
Thank you!

PR #2044, opened 2021-05-25
Add link to new Manning book

gruberb:
There is a new Manning Rust book in MEAP and I think it is interesting for the community to know!


nellshamrell (approved):
Ty!

PR #2056, opened 2021-05-27
Fixes broken Markdown formatting in 392

PTNobel:


jplatte:
The second link is still broken after this change. The bit you removed in the second commit needs to be inserted where you only inserted a closing parenthesis in the first commit.

nellshamrell (approved):
Ty!

PR #2057, opened 2021-05-27
Remove duplicate link in 392

Sosthene-Guedon:
The link [Why and how we wrote a compiler in Rust (blog post series 1/X): the context](https://bnjjj.medium.com/why-and-how-we-wrote-a-compiler-in-rust-blog-post-series-1-x-the-context-e2f83b10edb9) is present twice.


nellshamrell (approved):
Good catch! Ty!

PR #2067, opened 2021-06-02
Take Web Snapshot & Make Watermark in Rust

zzy:
Take web snapshot  from url, and make watermark for it.


nellshamrell (approved):
Thank you!

PR #2072, opened 2021-06-03
Remove leftover "Let's go full meta" from last week's QotW

joshtriplett:
Doesn't apply to this week's.


nellshamrell (approved):
Thank you!

PR #2074, opened 2021-06-03
Add blog post on Behavior inheritance in Rust

smokku:
This is a blog post from January, but I think it still is worth mentioning.


nellshamrell (approved):
Ty!

PR #2078, opened 2021-06-05
Fix Vancouver BC Rust meetup country abbreviation

jonathanrlouie:
The Vancouver B.C. Rust meetup is in Canada, not the U.S.


nellshamrell (approved):
Apologies and Ty!

PR #2083, opened 2021-06-09
Build front-end web apps with Yew and WebAssembly in Rust

zzy:
- Build front-end web apps with Yew and WebAssembly in Rust -part 1: crates
- Build front-end web apps with Yew and WebAssembly in Rust -part 2: Components & Routers


nellshamrell (approved):
Thank you!

PR #2084, opened 2021-06-09
Blog post on sleeping / power saving for embedded devices with Rust

erikjee:
This explores first steps to use sleeping for energy efficiency in embedded devices.

Not sure this should be in the walkthroughs category, but seemed the best guess :)


nellshamrell (approved):
This is great! Thank you!

PR #2090, opened 2021-06-09
Fix likely typo 🇺🇸 ⇒ 🇨🇦

scooter-dangle:


nellshamrell:
Heh, it was, indeed, a typo. Got it fixed in another pull request, but ty!

PR #2097, opened 2021-06-10
Add link to translation article

djc:
Uses a new Rust crate under the covers (which is mentioned in the article), is that on-topic enough?

(Note: the From Julia to Rust article seems to appear twice.)


nellshamrell (approved):
Ty! And I will fix the duplicate link, ty for bringing it my attention :)

PR #2115, opened 2021-06-17
Make content warning more prominent

jplatte:
`CW` doesn't seem to be that common of an abbreviation [based on reddit comments](https://www.reddit.com/r/rust/comments/o1l0xo/this_week_in_rust_395/h22i1mo/?utm_source=reddit&utm_medium=web2x&context=3). This commit writes it out, emphasizes it and puts it on a separate line (seems like it was on the same line as `**Highlight**` by accident).


nellshamrell (approved):
Quite reasonable! Thank you!

PR #2123, opened 2021-06-22
suggest new post in `Walking through "The Java Tutorials" with Rust`  to Observations/Thoughts section

aviillouz:
Walking through "The Java Tutorials" with Rust - Boxed trait objects and the search for inheritance


nellshamrell (approved):
Thank you!

PR #2150, opened 2021-07-06
suggest 'What Is an Interface?' and Rust's impl specialization

aviillouz:
suggest new post in Walking through "The Java Tutorials" with Rust - Reading 'What Is an Interface?' and trying Rust's impl specialization


nellshamrell (approved):
Thank you!

PR #2153, opened 2021-07-06
Add proposal for my article on closure traits

selectiveduplicate:
Proposes the article titled "Rust closures: How a closure can implement more than one trait" for the upcoming issue.


nellshamrell (changes requested):
Unfortunately, it looks like that link is broken (I get a 404 when I attempt to go to it). Do you mind fixing it?

selectiveduplicate:
> Unfortunately, it looks like that link is broken (I get a 404 when I attempt to go to it). Do you mind fixing it?

Fixed. Should be working now.
So sorry for that! My vim seems to insert some rubbish characters upon saving Markdown files for some reason.

PR #2172, opened 2021-07-13
update Rust description

crd477:
Make the brief blurb about the Rust language match the one from the official language page.

I actually don't feel strongly one way or another about this change but I figured I'd point it out in case it was a genuine oversight.


nellshamrell (approved):
Works for me! I appreciate this!

PR #2182, opened 2021-07-16
Add wgpu news to TWIR-400

kvark:
I'm not sure if "Franzplot" should be in any other category than "project updates".
Also the ordering is arbitrary, feel free to re-order.


nellshamrell (approved):
Ty!

PR #2186, opened 2021-07-19
Knurling-rs user survey

Urhengulas:
Hi,

We are not super sure if twir will accept community surveys, but we thought we will try. It could maybe even fit into the "Call for Participation" sections :D

Thank you for your work!

Best,
Johann


nellshamrell:
We do indeed! I'll move it to the CFP section. And you are very welcome, ty for your work too!

nellshamrell (approved):

PR #2198, opened 2021-07-20
Added walkthrough for rust in electron

domtac:
Noted the steps necessary to setup a wasm with wasm-pack in a electron app derived  from electron_boiler_plate project


nellshamrell (approved):
Ty!

PR #2220, opened 2021-07-27
Blog article on async and low power for embedded

erikjee:
Our research plan (kind of) on a test case for async / low-power with Rust embedded.


nellshamrell (approved):
Ty!

PR #2221, opened 2021-07-27
RTIC Monkeys, Guiter Hero on Rust Embedded

erikjee:
Not sure this belongs in TWiR, but it is fun and definitely falls in the "Miscellaneous" category :)


nellshamrell (commented):
Heh - certainly cool, but not quite enough Rust content to include in the newsletter. Thank you anyway, though!

PR #2236, opened 2021-07-29
Add connectorx

dovahcrow:
Connectorx is a library that accelerates data loading from databases to dataframes. It is written in Rust and provides a Python interface as well.


nellshamrell (approved):
Ty!

PR #2239, opened 2021-07-31
Adding Deep Learning in Rust with GPU

haixuanTao:
Deep learning in Rust with GPU talks about the hit and miss of doing Deep Learning with Rust.


nellshamrell (approved):
Ty!

PR #2244, opened 2021-08-03
Add "this week in sixtyFPS"

hunger:
I pushed this to reddit for the last couple of weeks and got encouraging
feedback there as well as on the rust user forums. So I thought it might
be interesting enough to push here:-)


nellshamrell (approved):
Ty!

PR #2252, opened 2021-08-04
Add tokio-internals zh-cn

pickfire:
Fix #2213

PR #2264, opened 2021-08-05
Mention that ZH for tokio internals is simplified

pickfire:
Some follow up for https://github.com/rust-lang/this-week-in-rust/pull/2252


nellshamrell:
Thank you so much for correcting this (and my apologies, I was the one who changed the designation). Merging and deploying immediately!

PR #2268, opened 2021-08-06
Fix a typo in 2021-08-04

teor2345:
Make the link title match the blog post name


nellshamrell (approved):
Good catch! Thank you!

PR #2319, opened 2021-08-20
Add meetup recording "Traits and trait objects - more than just interfaces"

FSMaxB:
I am not 100% certain if "miscellaneous" is the correct category.


nellshamrell (approved):
I think this fits better in walkthroughs - I will move it there :)

nellshamrell (approved):

PR #2355, opened 2021-08-30
Content suggestion

janos-r:
Content suggestion:
My short article about ECS in game engines. Why it is practical to know these concepts and how it can start your journey into game engines like Bevy.


nellshamrell (approved):
Ty!

PR #2356, opened 2021-08-30
fix "Daily Rust: Iterators" links

crd477:
fix the link in the first occurrence of this entry and remove the second occurrence, effectively a duplicate.


nellshamrell:
This one was fixed in a previous PR, but ty anyway!

PR #2360, opened 2021-08-31
2021-09-01: Add "How to avoid lifetimes annotations in Rust (and write clean code)

skerkour:
Hello,
I've added a post about "How to avoid lifetimes annotations in Rust" to the "Observations/Thoughts" section

Cheers ✌️

### DIFF
--- a/draft/2021-09-01-this-week-in-rust.md
+++ b/draft/2021-09-01-this-week-in-rust.md
@@ -20,6 +20,8 @@ In the case of this newsletter, 404 is indeed found!
 
 ### Observations/Thoughts
 
+* [How to avoid lifetimes annotations in Rust (and write clean code)](https://kerkour.com/blog/rust-avoid-lifetimes-smart-pointers/)
+
 ### Rust Walkthroughs
 
 ### Miscellaneous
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hi @skerkour!

I want to start off by saying how much I've enjoyed featuring your blog posts (I've learned a lot from them too!) and I'm looking forward to the general release of Black Hat Rust!

For this one specifically, particularly the beginning of the post, I feel it crosses into being a rant. Rants, particularly ones which are about things that other community members have worked on, are something I don't include in This Week in Rust. This is not to say your post does not have valid points - it certainly does and I think your tips on how to avoid lifetimes when you don't need them are quite valuable.

Would you be willing to re-word the beginning of the post so that it is less sarcastic and ranty? Or do another post (without the intro) that includes your information about how to avoid lifetimes when you don't need them? If so, I would be happy to feature it in next week's issue.

skerkour:
Hi @nellshamrell 
Thank you very much for the kind words and the feedback!

For sure I can rewrite the introduction, that I admit, is a little bit too much offensive.

I'll close this PR, take time to rewrite, and open a new PR for next week's issue :)

Kind regards,
Sylvain

nellshamrell:
Sounds great! Ty!

PR #2361, opened 2021-08-31
New Zellij release post

imsnif:
Hey,

We just released a new version of Zellij. This one is a major release so we wrote a blog post about some of the major features. We'd be very happy if you'd like to include it in the newsletter.

Thanks!


nellshamrell (approved):
Ty!

PR #2380, opened 2021-09-04
Remove "404" special message from issues 406 and 407.

### LINKED ISSUE #2376: 406 != 404
> In the case of this newsletter, 404 is indeed found!

This is two weeks old 🙂

kpreid:
Fixes #2376.


nellshamrell (approved):
Ty so much!

PR #2386, opened 2021-09-06
Add walkthrough video of setting up an Adruino Project using Rust

crazcalm:
Thank you for your consideration!!!


nellshamrell (approved):
Ty!

PR #2397, opened 2021-09-08
2021-09-08: Add How to avoid lifetime annotations in Rust

skerkour:
Hi,

I've rewritten the introduction of my post "How to avoid lifetime annotations in Rust (and write clean code)" which was too offensive.

I've removed all the exaggerations and opinions that may create polarization, to focus on outlining the real problem, which is, in my opinion, code clarity.

I'm making this PR as a separate one, so if the post needs more edits, the other one can still be accepted :)

Sylvain ✌️


nellshamrell (approved):
This is MUCH better! Thank you!

skerkour:
Thank you ;)

-------- Original Message --------

PR #2404, opened 2021-09-10
Fix typo: softwware -> software

matthiasbeyer:
The title of the [article](https://dpc.pw/data-oriented-cleanandhexagonal-architecture-software-in-rust-through-an-example) seems to have a typo in TWiR.


nellshamrell (approved):
Good catch! Ty!

PR #2424, opened 2021-09-17
URL to Playlist of "From Python to Rust" Series

mike-kfed:
Video series explaining how existing Python programmers can transfer their code over to Rust.

Please note that this is a shameless plug, as I am the author of those Videos! However I think this might be interesting to share with the community.


nellshamrell (approved):
Looks good! Ty!

PR #2434, opened 2021-09-22
2021-09-22: Add Rust on ESP32

skerkour:
Hello,

I've added two guides on how to run Rust on an ESP32 microcontroller to the Walkthroughs section

Cheers ✌️


nellshamrell (approved):
Ty!

PR #2449, opened 2021-09-28
Talk at GI

skade:


nellshamrell (changes requested):
Hi @skade! I only include events that will be happening in the next two weeks in the events section. However, I don't mind having this announcement in the newsletter in a different section. Do you mind moving it to the miscellaneous section?

PR #2456, opened 2021-09-29
Use the more accurate wording of 'Rust project' for the PR updates section

rylev:
This wording is a more accurate reflection of the terms used in the project since these are not changes from the core team but the project as a whole.


nellshamrell (approved):
Makes sense! Ty!

PR #2462, opened 2021-09-30
add link to SyRust paper

crd477:
added a link the SyRust paper.  I expect that Miscellaneous is the right section but feel free to move it to another if there's a better one.


nellshamrell (approved):
We haven't had a reason to have a "Research and Papers" section for awhile - this is a welcome one! :)

PR #2464, opened 2021-10-01
Adding Rust role from pganalyze

Vrq:


nellshamrell (commented):
For some reason that link is not rendering for me - do you mind double checking it? (I've tried in Firefox and in Chrome)

Vrq:
@nellshamrell looks like a temporary issue - should work now :)

nellshamrell (approved):
Looks good! Ty!

lfittl:
@Vrq I was quite surprised to find our 2 month old job posting in this week's newsletter.

Not sure if you did this intentionally or not (are you affiliated with"devitjobs.us"?), but this is an old job posting from August that is no longer active.

We have contacted DevITJobs since they copied this posting without our knowledge, and asked them to remove the post from their site. They have also posted their copy of the posting to other job boards, seemingly in an attempt to drive traffic to their site.

Whilst I appreciate that you promoted our job posting, this will likely leave candidates confused (as this position is no longer open), and I'd ask you to not repeat this for us.

@nellshamrell As a side note, we really appreciate it that you posted our job here, but it'd be great if the editors could do a little bit more due diligence here in the future (e.g. checking if the requester actually is affiliated with the company).

PR #2466, opened 2021-10-03
add guide on how to export mocks across crates

nrxus:
Adds a guide on how to export test mocks across crates. The guide itself focuses on the library `faux` but the solution is library-agnostic.


nellshamrell (approved):
Ty!

PR #2467, opened 2021-10-03
add article: Does the published crate match the upstream source?

ericseppanen:
I spent a few weeks researching the form of a published crate, and discovering that it can be quite difficult for many crates to find the exact released code in the upstream repository. I learned some interesting things along the way and wanted to share my findings.


nellshamrell (approved):
Ty!

PR #2470, opened 2021-10-05
Article on power consumption measurements on nRF52 async vs sync

erikjee:
Another Embedded Rust article, this one about comparing power consumption for async and sync. 
Observations/Thoughts seemed the best place to put it.
Thanks!


nellshamrell (approved):
Ty!

PR #2476, opened 2021-10-06
add a machine learning related project

kemingy:
[Mosec](https://github.com/mosecorg/mosec) is a machine learning model serving framework built with Rust & Python.

Rust tokio provides a high-performance service layer, Python makes it easy and flexible to use with the current machine learning ecosystem.

### DIFF
--- a/draft/2021-10-06-this-week-in-rust.md
+++ b/draft/2021-10-06-this-week-in-rust.md
@@ -35,6 +35,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ### Miscellaneous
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hello there! This certainly looks like a cool project! I definitely see that at least part of this project is written in Rust. However, the examples in the README seem to all be Python examples. Is this meant to be used with Rust or Python projects? If it's not meant to be used with Rust projects, could you write up a blog post highlighting how you are using Rust and what you learned about Rust in the process of writing this project? It needs some more Rust related content to be included in the newsletter.

kemingy:
Thanks for your advice. I'll write a blog about my experience with Rust in this project.

PR #2505, opened 2021-10-15
Move Rust Iran meetup to online meetups

hamidrezakp:
Rust Iran Meetup is an online event that will be held in Iran not North America, and should be in online section.


nellshamrell:
Sincere apologies! That was a mistake on my part!

PR #2560, opened 2021-11-01
Added article - Advanced Serde - Loose type deserialization for non-self-describing protocols

willemolding:
Full disclosure I am the author of this article. That said I think I think a lot of people will find it very interesting. I haven't seen any other examples of using serde in this way.

Keep up the great work!


nellshamrell (approved):
Ty!

nellshamrell:
And it's completely fine to submit your own articles :)

PR #2563, opened 2021-11-02
Update 2021-11-03-this-week-in-rust.md

eZanmoto:


nellshamrell (approved):
Ty! I will go ahead and publish all three of these in this week's issue, but in the future do you mind limiting it to two videos from the same series in one issue?

eZanmoto:
Oh, apologies, thanks for letting me know. It actually just happened that I published three on separate days this week, but yes, I can limit the uploads to 2 a week in future. Thanks!

PR #2564, opened 2021-11-02
Add "Anatomy of a Terminal Emulator" post

imsnif:
Hi! This is a post that explores how terminal emulators work, including some Rust code snippets to demonstrate their capabilities. Would be great to be included in the newsletter!


nellshamrell (approved):
This is great! Ty!

I particularly ❤️ "As a person who is sometimes prone to anxiety, I urge you: when in doubt, please err on the side of caution and kindness." I heartily agree!

PR #2565, opened 2021-11-02
Add "The Tree Structure of File Systems" blog post

fkohlgrueber:
I'm not sure if this post fits the criteria of TWIR, but I thought it was worth bringing it up. 

The post discusses how the tree structure available in common file systems is different from a simple tree structure you'd implement in a programming language. I'm using Rust for the code examples and use Rust's way of splitting modules into files as an example of how a different file system could feel more natural. I'm explaining how splitting modules into files work along the way too.

The post isn't about Rust primarily, but has quite some references to it. I think the topic might be interesting for a lot of readers of TWIR. That's why I'm proposing it here.

What do you think? Thanks in advance ;-)

EDIT: Not sure about the category. Could fit to "Miscellaneous" as well...

### DIFF
--- a/draft/2021-11-03-this-week-in-rust.md
+++ b/draft/2021-11-03-this-week-in-rust.md
@@ -25,6 +25,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ### Observations/Thoughts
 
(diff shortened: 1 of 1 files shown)


nellshamrell (approved):
This is great! Ty!

PR #2576, opened 2021-11-04
415: Remove duplicated entry

matthiasbeyer:
Found this entry to be duplicated, remove it!

---

If you scroll up just a little bit, you'll find that this entry is already there.


nellshamrell (approved):
Ty!

PR #2579, opened 2021-11-06
Add "What does `&mut &[T]` mean?" article

WaffleLapkin:
I'm not sure what category I should use, but hopefully it's ok :)


nellshamrell (approved):
Ty!

PR #2580, opened 2021-11-06
twir-415: use consistent tags for videos

WaffleLapkin:
It seems like videos are usually tagged with `[video]` at the start, and not `(video)` in the middle. Also, it seems that videos are usually placed below articles.


nellshamrell (approved):
Good catch! Ty!

PR #2583, opened 2021-11-09
Add "Automating Ember releases with Rust" to Observations/Thoughts section

locks:
Hi :)

I added a blog post I wrote, "Automating Ember releases with Rust" to the Observations/Thoughts section.
It is my first time contributing so I am not sure that is the correct place to put it?

Thanks for your time and let me know if I need to change anything!


nellshamrell (approved):
This is great! Ty!

PR #2585, opened 2021-11-09
add trailofbits blog entry

crd477:
add a reference to the "Write Rust lints without forking Clippy" blog post


nellshamrell (approved):
Ty so much!

PR #2589, opened 2021-11-10
add style guidelines to README

ericseppanen:
This is intended to capture the style decisions that are agreed on between the different editors.


nellshamrell (approved):
Ty so much!

PR #2590, opened 2021-11-10
2021-11-10: Add Benchmarking symmetric encryption (AEAD) in Rust

skerkour:
Hi,
I've added a post about Benchmarking symmetric encryption (AEAD) in Rust in the "Observations/Thoughts" section

Cheers :v:


ericseppanen (approved):

nellshamrell (approved):
Ty!

PR #2595, opened 2021-11-10
Add Rust Los Angeles events

ShakenCodes:
Putting in the next event and the one after.

BTW, do events that aren't past hang on from week to week? We added our December event, as well as next week's. Do we need to add this again next week?

Thank you for this awesome resource!


nellshamrell:
You are very welcome! We only include events that are occurring in the next two weeks, so I'll only include the November event in this issue. However, please feel free to add the December event closer to then :)

PR #2609, opened 2021-11-16
Update RFC & FCP sections for TWiR-417

U007D:


nellshamrell:
Looks great! Do you mind adding this directly to the draft of the newletter? That way I can just merge it in :)

nellshamrell:
Went ahead and copy and pasted this into the current draft :)

U007D:
I can do that each week, if you prefer.

PR #2611, opened 2021-11-16
Community updates for 2021-11-17

ericseppanen:
This is my first time composing the community PR, so feedback is encouraged!

Things I might have gotten wrong:
- I put the "Tag1 Joins the Rust Foundation" under the **Foundation** heading. It's not news *from* the Foundation, so maybe that's wrong?
- I chose to sort articles in the order [text, audio, video] in each sub-section.

Consider this PR a work in progress until Tuesday midnight PDT; I'll keep adding anything the shows up during the rest of the day.

PR #2612, opened 2021-11-16
Update 2021-11-17-this-week-in-rust.md

eZanmoto:


ericseppanen (approved):
I would suggest putting a space between the `[series]` and `[video]`, for consistency with PR #2611 (also because it makes the markdown syntax un-ambiguous).

nellshamrell (approved):
Ty!

PR #2614, opened 2021-11-17
2021-11-17: Add 8 Ways to backdoor a crate in Rust for fun and profit

skerkour:
Hi,
I've added a post about "8 Ways to backdoor a crate in Rust for fun and profit" in the "Observations/Thoughts" section

:v:


nellshamrell (approved):
Ty!

PR #2630, opened 2021-11-22
add link to Calling Rust from Python using PyO3

saidvandeklundert:
I wrote a blog on how you can use PyO3 to call Rust directly from Python and would like to add a link to the post to the "Rust Walkthroughs" section.


ericseppanen (approved):
Great article!

nellshamrell:
Ty!

PR #2631, opened 2021-11-23
update(links.txt): TWiR 417, 2021-11-17

hmble:
Shall I create new file for next TWIR issues ? Current `links.txt` file is now around 3000+ lines.

We can create a subfolder named `extracted-links` and then store files in that folder. As it wont clutter the root of project with txt files. This way we can have files of around 3000 lines per text file which makes it easy to scroll to check for duplicate links.


nellshamrell:
@hmble that sounds good to me :)

PR #2643, opened 2021-11-25
remove tracking parameters from link in 2021-11-24 issue

ericseppanen:
I think we have reached a consensus that tracking parameters on links are undesirable and should be removed.


nellshamrell (approved):
Good catch! Ty!

PR #2646, opened 2021-11-25
Use bullets for "Approved RFCs"

markmmm:
There are two entries - but they appear on one line making it a little less visible.
Turning it into a bulleted list to make it clear that there are 2 Approved RFC's.


nellshamrell (approved):
Ty!

PR #2665, opened 2021-12-02
Shouldn't Blackhat be in "Rust Walkthroughs'" section

luisvonmuller:
I think thats not the right fit for it (place to be), and must be in the **Rust Walkthroughs'** section. 

It's a project, for sure. But not such as a totally free project on rust (black hat rust - the book), neither it's a crate/tooling? - Its kinda an Article/How-to/Tutorial, so maybe: **Walktrought.."'s section then?


ericseppanen:
I suggested "Miscellaneous" in the original PR, but the submitter didn't go for it.

nellshamrell (approved):

PR #2672, opened 2021-12-06
remove articles that have copied heavily from other sources

ericseppanen:
Also fix the 2021-06-02 qotw to have correct attribution + link.


luisvonmuller:
If we start with stuff like this... we should also ban paywall/monetized content. I don't know if I really do want to enter this kind of topic (cuz we should firstly carefully guarantee that every one is actually OC, do we have time for this? Is it worth?). 

After all, the ones that does are doing, and I do think that this whats matters the most for the community.

nellshamrell:
We do ban paywall/monetized content (at least, it is listed in the [README](https://github.com/rust-lang/this-week-in-rust#readme) as something we are not looking for).

I see this in part as being a good community citizen - if someone copies someone's work, only changes it slightly, and then publishes it  under their name...that's not being a good community citizen. And that's not something that I want in This Week in Rust.

PR #2673, opened 2021-12-06
 Adding the Brazilian unofficial translation 🇧🇷 

luisvonmuller:
First things first, as always.
This translation idea came from getting people interested in The Rust programming language. And by that I mean there is no reason by now to translate things like RFC's, rust's PR's, English Events and others like (as well jobs, by now).

This is just a small step for the rust community, but **the** big step for the Brazilian's Rust Community. :)  

The right way maybe should be "PT-BR" - but I'll just change to this when Portugal give us our gold and "Pau Brasil" back.


ericseppanen:
`PT-BR` seems fairly common also. Whatever is most useful to the readers in Brazil is probably fine.

I also notice this commit is based on a different PR-- it would probably be better to rebase this PR on the master branch.

ericseppanen (approved):

nellshamrell (approved):

PR #2678, opened 2021-12-08
Update Events section for TWiR 420

mariannegoldin:
This PR:
* Adds events for Dec 1-15 2021
* Sorts events (in order of: type (online/physical), locale, date, city name, group name)
* Removes cancelled or inaccurate events


nellshamrell (approved):
Ty so much!

PR #2681, opened 2021-12-08
2021-12-08: Add Rust streaming file encryption

skerkour:
Hi,
I've added a post about how encrypt files in Rust using streaming AEAD encryption

Cheers
✌️


nellshamrell (approved):
Ty!

PR #2687, opened 2021-12-09
Add this week in Databend 19

PsiACE:
Ty!


nellshamrell (changes requested):
Since that issue has already been published, do you mind adding it to the draft of next week's issue?

PsiACE:
> Since that issue has already been published, do you mind adding it to the draft of next week's issue?

Next week, I will add a new weekly on time

PR #2690, opened 2021-12-09
Note non-video version of On Hubris And Humility

cbiffle:
OSFC doesn't caption or transcript their videos, so, linking the text transcript on my blog seems like it will improve accessibility.

Not sure if you'd rather note this in a future issue.


nellshamrell (approved):
Thank you so much for this! Adding this in before I send out the email version of the newsletter.

PR #2693, opened 2021-12-12
Add Minority Game Demo video

ecton:
Added a video showing building a networked application using two early-in-development crates we've been working on. The demo being built was for yesterday's Rust Gamedev Meetup.

I believe this was the best section to place it based on similar videos in previous newsletter entries.


nellshamrell (approved):
Ty!

PR #2701, opened 2021-12-15
Add events to TWiR #421

mariannegoldin:
This PR adds events for 12/15/2021-1/15/2022 (slightly longer period given the holiday break).


nellshamrell (approved):
Ty so much!

PR #2725, opened 2021-12-22
Update events for TWiR 422

mariannegoldin:
Deleting some old events and adding the Austin event in Jan.


nellshamrell (approved):
Ty so much!

PR #2733, opened 2021-12-24
add missing right guillemet for Chinese link in 2021-12-22

liubin:
The link's text missed a right guillemet,
the `《` and `》` should be in a pair.

Signed-off-by: bin liu <liubin0329@gmail.com>


nellshamrell (approved):
Ty!

PR #2736, opened 2021-12-25
Update links.txt

hmble:


hmble:
Extracting links to a a file doesnt seem to be effcient solution. Now that we have a script [tools/inspect_links.py](https://github.com/rust-lang/this-week-in-rust/blob/master/tools/inspect_links.py) I dont think its necessary to keep scraping links to txt file.

So for now I think we stop extracting links. 
@nellshamrell I would like to know your opinion about this.

nellshamrell:
@hmble I think that's fine. I really appreciate you doing this for so long!

nellshamrell (approved):

PR #2739, opened 2021-12-26
[walkthroughs] Parlons Peu, Parlons Rust!

o0Ignition0o:
Hey!

I'd like to add a link to a series of fr videos I've been creating.
In this series of episodes we're learning the basics of Rust, one step at a time. 5 episodes are available at the moment, and the 6th about structures should be out in two weeks.


nellshamrell (approved):
Très bien! Merci!

PR #2740, opened 2021-12-27
fix link markdown that's not parsed correctly

ericseppanen:
The previous formatting was just emitting a blank line; try a different formatting that seems to work.


nellshamrell (approved):
Ty!

PR #2756, opened 2021-12-30
Restore full content inside RSS feed

ThibaudDauce:
Hi,

Having all the links directly inside the feed allow me to add links to Pocket without leaving my RSS reader.

I don't know if I'm the only one with this workflow, but since there is no ad in the main website, I don't think there is a reason to force the reader to go to the main website to read the links?

This was the behavior before the update a few weeks ago.

Best regards,
Thibaud


nellshamrell (approved):
Thank you!

PR #2759, opened 2022-01-01
Update 2022-01-05-this-week-in-rust.md

CarlKCarlK:
Add free link to an article about Rayon (including capturing all errors and letting users control number of threads) and Python extensions. Based on experience with Bed-Reader bioinformatics project.


nellshamrell (approved):
Ty!

PR #2768, opened 2022-01-04
Add github action workflow for link-checker script

ericseppanen:
When a push or PR happens, run the link-checker script.
- look back through 25 most recent issues (in /draft and /content)
- warn only for errors that exist in the 5 most recent issues (to allow for correction PRs that arrive after an issue is published).

This also upgrades Markdown from 3.3.5 to 3.3.6, because 3.3.5 seems to have been yanked for non-security reasons, and the github dependency script refuses to install yanked versions.


nellshamrell (approved):
Fantastic! Ty so much!

PR #2777, opened 2022-01-06
inspect_links: all logging to stdout

ericseppanen:
In github actions, stdout and stderr are interleaved weirdly, making the output hard to read ([example](https://github.com/rust-lang/this-week-in-rust/runs/4722235297)). To fix this, don't log messages to both-- just use stdout for everything.


nellshamrell (approved):
ty!

PR #2783, opened 2022-01-08
add a tool for detecting unrecognized html tags and broken backticks

ericseppanen:
The markdown syntax allows un-escaped < > characters to be emitted in the html output. This is probably undesirable, as something like `Vec<T>` will be interpreted as a tag.

example output:
```txt
content/2020-03-17-this-week-in-rust.md: unrecognized tag t in "<t></t>"
content/2020-03-24-this-week-in-rust.md: unrecognized tag t in "<t></t>"
content/2020-05-12-this-week-in-rust.md: unrecognized tag cow in "<cow> for Box, Rc, and Arc</cow>"
content/2020-05-27-this-week-in-rust.md: unrecognized tag statuscode in "<statuscode></statuscode>"
content/2021-12-22-this-week-in-rust.md: unrecognized tag alias in "<alias>`</alias>"
draft/2022-01-12-this-week-in-rust.md: unrecognized tag t, in "<t, a="">, pronounced Veck-tah. 😛</t,>"
```

This addresses part of #2532, though there are lots of other ways to mess up markdown, so I'm not sure whether to mark that "fixed" or not.

This file imports some code from inspect_links, which might be a bad design; separating "CLI tool" from "library" code is left for a future commit.

PR #2784, opened 2022-01-08
fix some unescaped angle brackets and broken backtick pairs

ericseppanen:
These are incorrectly interpreted as html tags.

This commit only fixes bad tag-like brackets since 2020 as found by #2783.


ericseppanen:
Added fixes for broken backticks too.

nellshamrell (approved):
Thank you!

PR #2790, opened 2022-01-11
Update 2022-01-12-this-week-in-rust.md (#1)

kw-fn:


ericseppanen (commented):
Why does this change the title of an existing link?

Unless there's a compelling reason, I think the link text should match the article title.

kw-fn:
Yes, the project name has been updated, technically references to the legacy name are outdated. 

I think that starting from the recent update, modifying all current/future references makes sense, even if the blog content its self hasn't been modified. From the perspective a future reader it's the legacy name which would seem out of place.

ericseppanen:
Sorry, but the policy here is to have the link title match the original post. As long as the authors of the blog chose to leave their title alone, we will too.

ericseppanen:
Another note: the TWiR editors do collect links from most of the common sources. The OSX link and the Fyrox name change announcement are unique; the rest will be handled by #2794.

PR #2806, opened 2022-01-14
Use direct link for 1.58.0 testing

teor2345:
The current link goes to the list of blog posts, which is slightly confusing.


nellshamrell:
Got this fix in a previous pull request, but thank you anyway!

PR #2815, opened 2022-01-18
add lipstick to project updates

jrvidal:


nellshamrell:
This certainly looks cool, but I'd love if you could add a little more than the code. I'd love some narration about what the project is, why you created it, and what people can do with it. The code is great, but I like seeing it accompanied with some more narration when featuring it in TWiR.

jrvidal:
Gotcha, makes sense. I'll resubmit if/when I have some words to accompany it.

PR #2821, opened 2022-01-19
Added BonsaiDb Benchmark Post

ecton:
I wrote a post with some (what I think are) early but exciting benchmark results for our pure Rust database.


nellshamrell (approved):
Ty!

PR #2823, opened 2022-01-19
Add GStreamer Rust Bindings 0.18 release announcement

sdroege:


nellshamrell (changes requested):
Hello there!

Thanks for submitting this.

I believe these release notes are a little sparse to be included in This Week in Rust. Would you consider either fleshing them out more or writing up a blog post with more information on:
* how people can use this new version of the crate (especially what it allows them to do in Rust, code examples are always most welcome)
* what makes it different from previous versions
* what you learned about Rust in the process of writing it

sdroege:
Too late for this round of twir so let's try again for next release :)

PR #2827, opened 2022-01-19
2022-01-19: Add How to securely encrypt a file with an insecure password in Rust (using Streaming Encryption + Argon2)

skerkour:
Hi,
I've added a post about combining Streaming encryption + KDF to securely encrypt data with an insecure password.

Cheers ✌️


nellshamrell (approved):
Ty!

PR #2833, opened 2022-01-20
Update 2022-01-26-this-week-in-rust.md

rrrodzilla:
Hi, please consider my article for next week. Thank you!


ericseppanen (approved):
Good article, thanks!

I noticed one typo, where you write "version 1.06"-- I think this should be "version 1.0.6", right?

rrrodzilla:
Definitely a type.  Just pushed out the fix.  Thanks very much!

nellshamrell (approved):
Great work! Ty!

PR #2835, opened 2022-01-21
add link to lowering async await in rust

contrun:
This article walks through the process of how rust generates async/await code.

### DIFF
--- a/draft/2022-01-26-this-week-in-rust.md
+++ b/draft/2022-01-26-this-week-in-rust.md
@@ -22,6 +22,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ### Rust Walkthroughs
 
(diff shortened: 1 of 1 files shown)


ericseppanen:
Nice article!

I'm curious whether there's something broken with the page layout, though. On a desktop browser, all the content gets squeezed into a skinny column at the far left that's not wide enough for all of the code to fit without horizontal scrolling.

contrun:
Thanks.

The layout is actually a feature. First, it will be too wide if the content occupies the whole screen. That will be hard to read. Second, my theme is copied from https://braindump.jethro.dev/posts/. It can stack articles when you click on a internal link. Here is an example https://wiki.cont.run/self-hosted-services/?stackedNotes=..%2Fself-hosted-infrastructure%2F . It is a bit strange to have a scroll bar not in the far right, but it is by design. 😊

nellshamrell (approved):
Ty!

PR #2838, opened 2022-01-21
Add rustdoc-settings-sync link

notriddle:


ericseppanen:
TWiR doesn't usually publish links to source repositories.

If you were willing to write up a blog post somewhere, describing the problem, how you solved it, lessons for other rust users, etc., that would be something we'd likely accept.

notriddle:
Okay, I put together a blog post describing what the problem is, how I solved it, and a fair amount more background information.

nellshamrell (approved):
Thanks for being willing to make those changes :)

PR #2847, opened 2022-01-26
add sub-categories and link title policy to the README

ericseppanen:
Document the usual "community" sub-categories. I've tried to describe them as well as I understand the current conventions, which is certainly incomplete.

Also add a note about link titles, to document the current "leave page titles as-is" policy.


ericseppanen:
Reviewing this change, I wonder why "Foundation" is its own category at all. If you ask me, lumping Foundation blog posts under "official" would be fine, as neither of those topics are particularly crowded.

nellshamrell:
The reason we kept "Foundation" and "Official" separate was to separate the Foundation and the Project, since they are distinct entities. We don't want it to look like the Foundation's posts are the voice of the Project and vice versa.  Conflict between the messages of the two is unlikely, but I'd prefer to keep them separate for now.

PR #2883, opened 2022-02-02
Add FS <> AdaCore announcements

skade:
I hope "Miscellaneous" is the right category.


nellshamrell (approved):
I think Misc is the correct place :)

PR #2902, opened 2022-02-07
Add link to Why Not Inheritance

ifsheldon:
Add a link to a blog talking about why Rust does not have inheritance like other OO languages do.

### DIFF
--- a/draft/2022-02-09-this-week-in-rust.md
+++ b/draft/2022-02-09-this-week-in-rust.md
@@ -25,6 +25,7 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 ### Research
 
 ### Observations/Thoughts
+* [ZH] [为什么 Rust 没有继承？](https://fengliang.io/RustWHY/design_choices/why_not_inheritance.html) A thought on the design choice that Rust made and is questioned by many programmers from other object-oriented languages. Why does Rust not have inheritance?
 
 ### Rust Walkthroughs
 
--- a/draft/2022-02-09-this-week-in-rust.md
+++ b/draft/2022-02-09-this-week-in-rust.md
@@ -25,7 +25,7 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 ### Research
 
 ### Observations/Thoughts
-* [ZH] [为什么 Rust 没有继承？](https://fengliang.io/RustWHY/design_choices/why_not_inheritance.html) A thought on the design choice that Rust made and is questioned by many programmers from other object-oriented languages. Why does Rust not have inheritance?
+* [ZH] [为什么 Rust 没有继承？Why doesn't Rust have inheritance?](https://fengliang.io/RustWHY/design_choices/why_not_inheritance.html)
 
 ### Rust Walkthroughs
 


ericseppanen:
I don't think the text after the link is part of the TWiR style; I'd suggest removing that.

ericseppanen:
Can @Folyd or @MATRIXKOO review this?

ifsheldon:
It's just a little pitch. My intention is to give a brief before people consider reading it. If it's not appropriate, I'm fine with removing it.

Folyd:
Hi @ifsheldon, thanks for your PR. Just my two cents: 

- **audiences first, not you**.

For example, this sentence "*所以终极的原因还是代 cheng 码 xu 可 yuan 复 tou 用 lan*" brings no benefits to audiences except distractions. Personally, I don't think it is a good style for a technical article.

- **Dive deeper**.

For example, I'll recommend mentioning the `Deref` trait which can simulate the inheritance, and also should point out that is an [anti-pattern](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html).

Do you mind changing the writing styles and supplementing more detail about Rust's inheritance? That could be better for the audience. Thanks.

However, I also recommend [Rust Chinese Magazine](https://rustmagazine.github.io/rust_magazine_2022/), which is a better place to meet your Chinese audiences. Feel free to contribute to that. Here is the repo: https://github.com/RustMagazine/rust_magazine_2022

ifsheldon:
> * **audiences first, not you**.
> 
> For example, this sentence "_所以终极的原因还是代 cheng 码 xu 可 yuan 复 tou 用 lan_" brings no benefits to audiences except distractions. Personally, I don't think it is a good style for a technical article.

Why so serious? Articles can be fun and at the meantime rigorous. But yeah I agree that this line is not necessary, so I changed it.

> * **Dive deeper**.
> For example, I'll recommend mentioning the `Deref` trait which can simulate the inheritance, and also should point out that is an [anti-pattern](https://rust-unofficial.github.io/patterns/anti_patterns/deref.html).

I don't think this article is not deep enough. `Deref` is just a technical detail but it does not mean it's deeper. From my personal perspective, I like Rust not because it's technically deep but it's conceptually elegant. My article resonates with Rust's design in that behind technical details, there are some deep introspections, like the introspection on "Why and Why not inheritance". It's not about how to implement inheritance technically but the design from the introspections, which makes Rust distinguishable. But sure, why not? We can have both, so, I added this anti-pattern to the appendix.

draft/2022-02-09-this-week-in-rust.md
```diff
@@ -25,6 +25,7 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 ### Research
 
 ### Observations/Thoughts
+* [ZH] [为什么 Rust 没有继承？](https://fengliang.io/RustWHY/design_choices/why_not_inheritance.html) A thought on the design choice that Rust made and is questioned by many programmers from other object-oriented languages. Why does Rust not have inheritance?
```
Folyd, line 28: > I don't think the text after the link is part of the TWiR style; I'd suggest removing that.

Agree.

ifsheldon: Fine, I have no interest in arguing more on this.

Folyd (approved):

nellshamrell:
Thank you @ifsheldon, @Folyd, and @ericseppanen!

PR #2903, opened 2022-02-08
Add BonsaiDb v0.1.0 announcement

ecton:
Also, thank you for picking up my January update in the last edition!


nellshamrell (approved):
Ty!

PR #2915, opened 2022-02-09
Adding Slight's Rust software engineer role

ColmanHumphrey:
Thanks again for maintaining this!


nellshamrell (approved):
"For the technical screen and the interviews, we will send all questions approximately a day in advance."

❤️ ❤️ ❤️  - that is such a healthier approach to interviewing than I see at most places!

ColmanHumphrey:
> "For the technical screen and the interviews, we will send all questions approximately a day in advance."
> 
> ❤️ ❤️ ❤️ - that is such a healthier approach to interviewing than I see at most places!

It's been good so far! Also we get higher quality answers on average this way too, which as an interviewer is also a plus.

PR #2926, opened 2022-02-12
Add link to Why Associated Type

ifsheldon:
Add link to Why Associated Type, which is my thought on the design of Associated Type. It's not about How, but Why, so I'd rather like to place it under Thoughts, not Walkthroughs.

### DIFF
--- a/draft/2022-02-16-this-week-in-rust.md
+++ b/draft/2022-02-16-this-week-in-rust.md
@@ -23,6 +23,7 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 ### Newsletters
 
 ### Observations/Thoughts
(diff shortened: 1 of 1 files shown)


nellshamrell:
@MATRIXKOO, @Folyd could you help me review this one?

Folyd:
Hi, @ifsheldon. Thanks for the PR. I strongly recommend you submit one PR of **Rust Why** after the book is finished instead of multiple PRs each with a single article.

ifsheldon:
I don't think that is gonna work. I still have multiple articles to write, some of which may not be submitted to TWiR. And a bundle of articles is just too long to read. The article is self-contained, so there's no context tightly related to the book/ article collection itself.

Folyd (approved):

nellshamrell (approved):
Ty!

PR #2928, opened 2022-02-13
Walkthrough about using wgpu-rs for image filters

redwarp:
I just published this blog post, it's a walkthrough about using wgpu-rs and compute shaders to do some image filtering.
In it, I explain step by step how to create a grayscale version of an image, using your GPU.


nellshamrell (approved):
Ty!

PR #2934, opened 2022-02-14
Update walkthroughs with Rust Embedded videos

keadwen:


ericseppanen (commented):
Please use the canonical `youtube.com` URLs, rather than the `youtu.be` ones (that redirect).

nellshamrell (changes requested):
Once the links are updated (as requested by @ericseppanen) I would be happy to merge these!

nellshamrell (approved):
Ty!

PR #2937, opened 2022-02-15
Walkthrough on implementing a custom async runtime

DomWilliams0:
This post is a devlog about implementing an async runtime in my game engine, and mostly focusses on the Rust-side of things rather than anything gamedev or game engine specific.

The original title is "Devlog #&#x2060;7: custom async runtime" but I think it should be tweaked to add more context.

Thanks!


nellshamrell (approved):
Ty!

PR #2944, opened 2022-02-16
Added the Updates from the Tinyverse newsletter

Michael-F-Bryan:


nellshamrell:
While this certainly looks cool and I know Tinyverse does things with Rust, I don't see enough Rust specific content to include this newsletter in our newsletter. Thank you for your submission, though, and should you have content that is more Rust specific we would welcome additional submissions!

Michael-F-Bryan:
Thanks for considering this, @nellshamrell. The underlying project is mostly built on Rust, but I'll agree that newsletter focuses more on community updates and doesn't mention it at all :sweat_smile:

PR #2951, opened 2022-02-17
Add cargo-binstall 0.6 announce

passcod:


nellshamrell:
Unfortunately, there is not enough content in here for me to include it in This Week in Rust. When it comes to project updates, I'm looking for examples of how to use the project in Rust, what you learned about Rust through updating the project, etc. Thank you for submitting this, though, and I hope to see more submissions from you in the future!

PR #2954, opened 2022-02-17
remove duplicate link; fix typo in link text

ericseppanen:
Twice duplicate links weren't detected because they used "http" instead of "https".


nellshamrell (approved):
Ty so much!

PR #2955, opened 2022-02-17
improve duplicate link detection

ericseppanen:
When checking for duplicate links:
- Assume http and https links to be the same.
- Ignore consecutive slashes and trailing slashes.

Consecutive slashes in the URL will trigger a warning; trailing slashes or http links will not.


nellshamrell (approved):
Ty!

PR #2956, opened 2022-02-17
Add recent nix performance improvements

SUPERCILEX:


ericseppanen (commented):
I think this is a really interesting discussion, but I don't think TWiR usually publishes links to a github issue discussion. I think this discussion is particularly hairy because the early code flirted with UB while trying to achieve performance.

Maybe a better way to publicize this would be to write up a blog post explaining the issue and the UB hazards, and how the code evolved to address those concerns. It's also not clear to me whether the 30x improvement was still upheld after the PR was rewritten using a different strategy.

SUPERCILEX:
> It's also not clear to me whether the 30x improvement was still upheld after the PR was rewritten using a different strategy.

Yup, the performance improvement is still there.

Have links from [Ralf's blog](https://www.ralfj.de/blog) been posted before? Anything there is amazing, though in this case https://www.ralfj.de/blog/2020/07/15/unused-data.html and https://www.ralfj.de/blog/2019/07/14/uninit.html are relevant. Maybe it makes sense to put those in the newsletter instead? I was more just trying to promote the nix performance improvement, so I don't think it makes sense for me to write a blog post about it. :)

ericseppanen:
Many of Ralf's articles have been posted to TWiR (including the unused-data link); we don't re-post old articles though.

I think that the world needs more writeups of actual real-world issues in real programs. The potential for UB makes this one really exciting, so I would still love to read the story if anyone would be interested in writing it up as a standalone article.

SUPERCILEX:
Alright, maybe I'll write something. :grin: In the meantime, closing.

PR #2957, opened 2022-02-17
Update 2022-02-23-this-week-in-rust.md

andrieshiemstra:
Added my walkthrough of embedding quickjs in a rust project

### DIFF
--- a/draft/2022-02-23-this-week-in-rust.md
+++ b/draft/2022-02-23-this-week-in-rust.md
@@ -18,6 +18,8 @@ If you find any errors in this week's issue, [please submit a PR](https://github
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


nellshamrell (changes requested):
This seems more like a Walkthrough than a project update, do you mind moving it to the "Rust Walkthroughs" section? And we normally don't include descriptions when we are listing articles, how would you feel about using the title "Embedding a JavaScript Engine in a Rust Project" for the link? That way it's nice and descriptive about the content.

andrieshiemstra:
I updated the link as requested, only added (quickjs) for clarity on what it is :)

Kind regards
Andries

nellshamrell (approved):
Ty!

PR #2973, opened 2022-02-22
C/QotW + notable changes

llogiq:


ericseppanen:
The CI failure is because the python markdown module interprets this as an html tag:
```
* [fix ICE when using Box<T, A> with pointer sized A](https://github.com/rust-lang/rust/pull/94043)
                         ^^^^^^
```
I suggest backticks to fix the problem:
```
* [fix ICE when using `Box<T, A>` with pointer sized A](https://github.com/rust-lang/rust/pull/94043)
```

nellshamrell:
Ty!

PR #2975, opened 2022-02-22
remove outdated and obsolete files

extrawurst:
the files were not updated in 2 months. the ci uses the md files directly for dup detection anyway


nellshamrell:
We decided to stop updating these after the CI was introduced. Merging.

PR #2976, opened 2022-02-23
community updates for 2022-02-23

ericseppanen:
Also move the ogma link as it's more of a walkthrough than a project update.


nellshamrell (approved):
Ty so much!

PR #2981, opened 2022-02-24
Correct date for Rust Hack'n'Learn

badboy:
Not sure how/where you got March 2nd.
It's Tuesdays, thus March 1st.


nellshamrell (approved):
Ty!

PR #2995, opened 2022-03-01
Add Rust runtime for AWS Lambda release announcement

calavera:


ericseppanen:
I don't think the TWiR style is to bundle two different links in the same line, but I'll let @nellshamrell decide what to do with this.

I also notice that there are unnecessary parameters on the twitter link-- I expect that the parameters can be safely removed (seems to work for me without them).

ericseppanen (approved):

PR #3002, opened 2022-03-02
2022-03-02: Add Black Hat Rust Cyberwar Discount

ghost:


ericseppanen:
I think we'll probably want to put this somewhere other than Project/Tooling Updates.

I'll let @nellshamrell decide.

nellshamrell (commented):
I agree on it not belonging in Project/Tooling Updates - going to move it to Miscellaneous (it is cool, though!)

PR #3027, opened 2022-03-10
Fixed copy/paste error in Approved RFCs section

U007D:


ericseppanen:
Thanks for noticing this! Since this is already published, I'll apply the fix to `content/2022-03-09-this-week-in-rust.md` instead.

ericseppanen:
@U007D can you confirm the form of the fix is correct?

PR #3040, opened 2022-03-15
add link to Rustacean Station ep058 transcript

ericseppanen:
This is a link to a text transcript of the Rustacean Station episode "What's New in Rust 1.58 and 1.59".

While this is the same content as the audio podcast, I'd like to call attention to the text transcript, as it's more accessible to those with limited English skills.

I'm certain that distinction is important enough to justify inclusion, so I'll leave this PR for @nellshamrell as the final word.


nellshamrell (approved):
Ty!

PR #3041, opened 2022-03-15
community updates issue 434

extrawurst:


ericseppanen:
A late arrival; should probably go in Observations/Thoughts:
```
* [Async IO with completion-model IO systems](https://www.ncameron.org/blog/async-io-with-completion-model-io-systems/)
```

I'm not sure what the best way is to handle late arrivals, considering we're in different timezones. Posting another PR just causes a merge conflict.
Two things to consider in the future:
1. Maybe we should do the community PR as a branch on the main repo? Then others could add stuff, rebase, etc. and the PR would automatically be updated.
2. Or maybe the right thing to do would be to just merge the community PR right away and open another one?

extrawurst:
@ericseppanen added the late addition

ericseppanen (approved):

PR #3055, opened 2022-03-21
Add RRC article to Thoughts/Observations

doc-jones:
Rustc Reading Club Phase I Retrospective link added under Thoughts/Observations


ericseppanen (approved):
Thanks!

PR #3072, opened 2022-03-24
Fix the spelling of "rust-analyzer"

jonas-schievink:


ericseppanen:
This repo contains 58 instances of "Rust Analyzer" and 10 instances of "Rust-Analyzer" (though some of them are in the titles of others' blog posts.)

@nellshamrell, in a case like this is it OK to do mass changes to past issues? Would it be preferable to just adopt the preferred name going forward?

nellshamrell:
@ericseppanen let's just adopt the preferred name going forward.

nellshamrell:
@jonas-schievink we will be sure to spell it in the preferred way going forward! Going to go ahead and close this PR, but thank you for bringing this to our attention :)

PR #3073, opened 2022-03-25
Add NXLog job

nxyury:


nellshamrell:
I'm getting a 403 when following that link - can you double check it?

pravic:
Works for me, for example.

PR #3075, opened 2022-03-28
Add "My First Clippy Lint" post

jamesmcm:
This post is a write-up of the process of writing my first Clippy lint for derecting the use of self as Display within the Display impl itself.


ericseppanen:
Thanks! This is a great article.

Does the `#en` on the URL do anything, or can we remove it?

jamesmcm:
Hey, it can be removed actually, it's just added by default for the unused multi-language support.

PR #3089, opened 2022-03-30
Add my GitHub profile generator blog post

autarch:
The profile generator is in Rust. I put this in Observations/Thoughts because I don't walk through the source code in the post.


ericseppanen (approved):
Thanks!

PR #3105, opened 2022-04-05
Add link to Coppers

ThijsRay:


ericseppanen:
Thanks for the links. The general style of TWiR is one link per line with the link text the same as the article title, and we prefer links to articles rather than repositories. Can I suggest this format?

```
* [Coppers - A Rust Test Framework That Measures Energy Consumption Over Time](https://raw.githubusercontent.com/ThijsRay/coppers/main/report.pdf)
```

I think this belongs in the "Research" section. The `### Research` title is missing from this week's draft, sorry for the oversight!

ThijsRay:
Thank you for the feedback!

I'm not really sure whether this would fall into the research category. The project has been developed in an academic setting and comes with a report for a bit of context, but it don't think it would qualify as research. Is it alright if I put in in the Miscellaneous section?

nellshamrell (approved):
Ty!

PR #3124, opened 2022-04-08
add a script to create the issue draft

ericseppanen:
This creates a draft issue from a template file, filling in basic things like the issue number, date, etc.

The script will create a new file in `draft/` unless the `--dry-run` flag is specified. It will fail if the file already exists.

I left html comments in a few places to mark where the content should go. Some of the comments are directed at contributors; feedback on this strategy is welcome as "how we communicate to contributors" is a separate issue from "how we create the draft file".

None of the sections are populated with content from the previous issue; I assume this would be desirable in some cases (events, probably?) but feedback on how to do it would be welcome. Maybe just copy/paste everything under `## Upcoming Events`?

### DIFF
--- a/tools/DRAFT_TEMPLATE
+++ b/tools/DRAFT_TEMPLATE
@@ -1,0 +1,138 @@
+Title: This Week in Rust $twir_issue_number
+Number: $twir_issue_number
+Date: $twir_issue_date
+Category: This Week in Rust
+
+Hello and welcome to another issue of *This Week in Rust*!
+[Rust](https://www.rust-lang.org/) is a programming language empowering everyone to build reliable and efficient software.
+This is a weekly summary of its progress and community.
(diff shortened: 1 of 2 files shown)


andrewpollack:
Thanks for making this! 
RE: `None of the sections are populated with content from the previous issue; I assume this would be desirable in some cases (events, probably?)` -- while copying all the events from the previous issue would be a niceity, it's not a dealbreaker if that can't be done. Best way would be to copy/paste everything between the `### Virtual` line, all the way to the `[community]: mailto:community-team@rust-lang.org` line

nellshamrell:
Hi @ericseppanen! I just tried this out generating the draft for next week's issue - only thing I noticed is that it did not delete the link under "Approved RFCs". Everything else worked great!

PR #3143, opened 2022-04-14
Add "digs - dig many at once"

azzamsa:


nellshamrell:
Hi @azzamsa! Thank you for submitting this, but we need more context to include this in This Week in Rust. Would you be willing to write up a blog post about why this crate was created and what you learned about Rust in the process? I would love to include something like that!

azzamsa:
Thanks for the heads up. 
Will do :)

PR #3172, opened 2022-04-22
add risingwave blog

wyhyhyhyh:
Added the blog describing RisingWave (https://github.com/singularity-data/risingwave), a cloud-native streaming database written in Rust.


ericseppanen (commented):
I don't think this post is suitable for This Week in Rust, as it doesn't contain any Rust content.

We would welcome future posts that have more Rust discussion to them!

PR #3176, opened 2022-04-24
Add blog post on self referential data structures

panstromek:
Adding my blog post on how first class self-referential structures could look like.


ericseppanen (approved):
Thanks!

PR #3182, opened 2022-04-26
Update template for RFC & FCP sections to new format

U007D:
It's easier (and less error-prone--ask me how!) to replace a single-line comment with the RFC & FCP content each week.  Thanks @ericseppanen for the template upgrade!

### DIFF
--- a/tools/DRAFT_TEMPLATE
+++ b/tools/DRAFT_TEMPLATE
@@ -70,25 +70,7 @@ If you are a Rust project owner and are looking for contributors, please submit 
 
 <!-- Perf results go here -->
 
-### [Approved RFCs](https://github.com/rust-lang/rfcs/commits/master)
-
-Changes to Rust follow the Rust [RFC (request for comments) process](https://github.com/rust-lang/rfcs#rust-rfcs). These
-are the RFCs that were approved for implementation this week:
-
-<!-- Approved RFCs go here -->
-
-### Final Comment Period
-
-Every week [the team](https://www.rust-lang.org/team.html) announces the
-'final comment period' for RFCs and key PRs which are reaching a
-decision. Express your opinions now.
-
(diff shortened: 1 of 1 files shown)


ericseppanen (commented):
By removing the boilerplate text, and titles, we would be reducing the value of a template file, because there's no place in the repo for editors to examine or improve that text.

I don't object to this PR, but just want to have the conversation in the open.  How useful is it to have the boilerplate text & formatting captured in the git repo?

nellshamrell:
@U007D what are your thoughts on @ericseppanen's comment ^?

ericseppanen:
On further reflection, I think "put all the boilerplate text in the template" is probably not worth the effort.

Maybe instead we should try to get to "commit all scripts that generate content into git"? That way at least if someone wants to update the boilerplate they can 'git grep' their way to success.

U007D:
Motivation:

I actually had an error in my submission one week because I had accidentally left the boilerplate header in the the file and then pasted the entire section (including the boilerplate header) in for that week.

Github "helpfully" showed the changed areas in the PR, but collapsed the now-duplicate header, so it was not visible upon review.

When the TWiR went live, I saw the problem and had to post a "hotfix" MR to remove the duplication.

> On further reflection, I think "all the boilerplate text must be in the template" is probably not worth the effort.
 I agree.  The goal here is not to "preserve the value of the template", but rather to have *something* which helps the editors in the publishing process.  Without boilerplate, the pasting process is much simpler (even *I* can't mess it up!) and more resistant to errors.
 
 If that is not compelling, there is still "value" to the template as it provides for consistent section ordering from week to week, and reduces merge conflicts.
 
 I'm interested to hear your thoughts on this.  Thanks, @ericseppanen.

ericseppanen:
I think we're in agreement.

U007D:
Oddly, TWiR #442's template came out *without* this change; I assume it's some sort of latency issue; will track it down if issue persists for TWiR #443.

PR #3191, opened 2022-04-27
create_draft: copy events from previous issue

### LINKED ISSUE #3171: create_draft: copy some content from previous issues?
The script `tools/create_draft.py` writes the new draft issue, with no content.

For some sections, it might be useful to copy over some content from the previous issue.

ericseppanen:
The create_draft script will now search for a previous issue (under draft and content) and then search that file for a range of lines that look like the events list, and copy those lines over to the new draft file.

If it has trouble finding the previous issue file, or can't find the events list, it will print a warning and then leave a placeholder comment in the draft issue where the events list should go.

Fixes #3171.


nellshamrell (approved):
Ty!

PR #3194, opened 2022-04-28
Add rust malaysia upcoming meetup

pickfire:


nellshamrell:
When I click on the link for the form I see the message "The form Rust Meetup May 2022 is no longer accepting responses.
Try contacting the owner of the form if you think this is a mistake."

pickfire:
Sorry, I forgot to press accept response but it should work now.

PR #3196, opened 2022-04-28
Remove 2 redundant links

jihchi:


andrewpollack:
Good catch, and apologies for the redundant links. This issue has been published, and work on it is archived. 

The next issue will not have these redundant links.

andrewpollack:
@ericseppanen Ahh, I hadn't known that, thank you for this information! I'll merge this in

PR #3216, opened 2022-05-04
Add this week in Databend 40

PsiACE:
Ty!

---

We plan to refactor our testing solution to introduce an extended version of [sqllogictest](https://www.sqlite.org/sqllogictest/doc/trunk/about.wiki). The statement spec could be categorized to following fields:

- `statement ok`: the sql statement is correct and the output is expected success.
- `statement error <error regex>`: the sql statement output is expected error.
- `statement query <desired_query_schema_type> <options> <labels>`: the sql statement output is expected success with desired result.

You can check out [RFC | New SQL Logic Test Framework](https://databend.rs/doc/contributing/rfcs/new_sql_logic_test_framework) to learn more.


nellshamrell (approved):
Ty!

Issue #3220, opened 2022-05-05
new Compiler Performance table does not render on site


fbstj:
in [twir #441](https://this-week-in-rust.org/blog/2022/05/04/this-week-in-rust-441/) the "new" compiler performance table introduced recently doe not render as a table.

![image](https://user-images.githubusercontent.com/19763/166888331-8d46a26e-73e4-420d-a0f4-647607b97844.png)

the table was added in #3208

ericseppanen:
Fixed by #3221.

PR #3221, opened 2022-05-05
Try to get the performance triage table to render correctly

ericseppanen:
I think the rendering problems in #3220 are due to a quirk of the python Markdown implementation, and I think this can be avoided by putting a blank line before the table begins.

@nellshamrell I'm not 100% confident that this fixes the problem, but I am confident it won't make anything worse.... would it be possible to just deploy this change to the site and see if the table rendering is fixed?

### DIFF
--- a/content/2022-05-04-this-week-in-rust.md
+++ b/content/2022-05-04-this-week-in-rust.md
@@ -134,6 +134,7 @@ Triage done by **@simulacrum**.
 Revision range: [1c988cfa..468492](https://perf.rust-lang.org/?start=1c988cfa0b7f4d3bc5b1cb40dc5002f5adbfb9ad&end=468492c2af3993f18b1fe98052200575c4a2e678&absolute=false&stat=instructions%3Au)
 
 **Summary**:
+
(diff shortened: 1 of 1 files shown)


ericseppanen:
For future reference, I was able to reproduce the original problem by manually rendering the markdown source in python like this:

```
md_text = open(filename).read()
html = markdown.markdown(md_text, extensions=['tables'])
```

ericseppanen:
I [experimented with Babelmark](https://babelmark.github.io/?text=**My+Table**%3A%0A%7C+Foo+%7C+Bar+%7C+Baz+%7C%0A%7C%3A---%3A%7C%3A---%3A%7C%3A---%3A%7C%0A%7C++1++%7C++2++%7C++3++%7C) a bit, and I guess I shouldn't have described this as a quirk of Python-Markdown; most markdown implementations require a blank line before the start of a table.

The quirk is that a few implementations *will* render a table without the blank line, and then only when certain syntax is present, like the table-header dashes...

Aloso:
@ericseppanen I confirmed that the blank line fixes the issue.

Mark-Simulacrum (approved):
Seems ok - a change to the upstream template in rust-lang/rustc-perf would also be great.

I'll hold off on merging since not sure how that works precisely in this repo.

nellshamrell (approved):
Ty!

PR #3223, opened 2022-05-05
Make tables more beautiful

Aloso:
Once the table markdown is fixed (#3221), tables are rendered like this with this PR:

bobby-tables


ericseppanen:
As far as I can tell, there are no markdown tables in any previous issues, so :+1: .

nellshamrell (approved):
Thank you so much!

Issue #3236, opened 2022-05-10
New "Acceptance Testing" section for RFCs?


epage:
Some times it can be a challenge to get testers for an RFC implementation to get the feedback to know if its ready for stablization (see [Mechanism for beta testing unstable features](https://internals.rust-lang.org/t/mechanism-for-beta-testing-unstable-features/14280)).

What if we had a new section "Acceptance Testing" where we queued up RFCs that are are at stage of needing general feedback, either as part of an intermediate design milestone (e.g. https://github.com/rust-lang/rfcs/pull/3243#) or in prep for stablization (e.g. https://github.com/rust-lang/cargo/issues/8415#issuecomment-1112618913)?  They'd need to link out to a test plan (defining what you are looking, link to how to use the feature, link to how to use it with nightly, where to apply feedback, etc).

jonhoo:
In the interest of continuing to leave a trail for things related to this effort, see also [The case for a new release channel: testing](https://internals.rust-lang.org/t/the-case-for-a-new-relese-channel-testing/14412/32), [Survey of "High Impact" Features](https://internals.rust-lang.org/t/survey-of-high-impact-features/14536/20), [RFC: Preview for Unstable Features](https://github.com/rust-lang/rfcs/pull/3120), and ultimately [this comment](https://github.com/rust-lang/rfcs/pull/3120#issuecomment-836921374).

I love the idea of a "call for testing" in TWiR. My main concern continues to be the ability to get widespread testing when it requires nightly.

U007D:
@epage Interesting thoughts.  Would it help if the section were titled something more "call-to-action" oriented e.g. "Call for Testing"?
@jonhoo I read the links you provided (only most of the RFC: Preview for Unstable Features) + your linked comment.  I have personal opinions on where we should land, but before I get to those, to address @epage's concern, it would be a shame to let perfect be the enemy of good.

To that end, a section with a call to action and clear instructions (or a link to clear instructions) that would evolve as the feature stabilization process does would allow us to move forward (if we elected to) quickly, while still preserving the option to evolve the stabilization process.

Definitely something to discuss with @nellshamrell and the rest of the TWiR editors, but if either of you have a draft of what you'd like to see the section look like please share it here, otherwise let us know.  I'm happy bring up the topic for discussion with the other editors.

epage:
Draft: comes before "Final Comment Period"
```md
### Call For Testing

[...]

PR #3249, opened 2022-05-12
Add section titled Call For Testing

Muscraft:
It was suggested in #3236 to add a section titled "Call For Testing". This was suggested as a way to increase the number of people testing "Pre-Stabilization" features. By adding this section there is a higher chance that people will see a feature is ready for testing and try it out.

Where this is meant section is meant to be placed can be moved around but it [was suggested](https://github.com/rust-lang/this-week-in-rust/issues/3236#issuecomment-1124144309) to put it before "Final Comment Period". As for the overall design of this section, I tried to make it match as best as I could while still making it readable. I am happy to tweak it to fit better if there are issues.

### DIFF
--- a/draft/2022-05-18-this-week-in-rust.md
+++ b/draft/2022-05-18-this-week-in-rust.md
@@ -70,6 +70,14 @@ If you are a Rust project owner and are looking for contributors, please submit 
 
 <!-- Perf results go here -->
 
+### Call for Testing
+
+An important step for RFC implementation is for people to experiment with the
+implementation and give feedback, especially before stabilization.  The following
+RFCs are at point where user testing is needed before moving forward:
+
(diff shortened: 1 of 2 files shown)


ericseppanen:
If you want some of this text to show up in *every* draft issue, please add it to `tools/DRAFT_TEMPLATE`.

Perhaps we should move the draft template file to the `draft` directory to make it more obvious?

Muscraft:
Moving it to the `draft` directory may be a good idea. I mostly just missed where I should be adding this. I was also trying to add the new section and the content for it and that caused issues. I'll open the PR for the content of this once we get this one merged.

nellshamrell (approved):
Ty!

U007D:
FYI: `Call for Testing` section language tweak in PR#3258.

PR #3258, opened 2022-05-17
Language tweak in new `Call for Testing` section

U007D:
Tweaked the language of the last sentence to avoid implying that the feature development would block on user testing.


epage (approved):

nellshamrell (approved):
Ty!

PR #3260, opened 2022-05-17
Update CFT, RFC and FCP sections for TWiR-443

U007D:


draft/2022-05-18-this-week-in-rust.md
```diff
@@ -72,15 +72,43 @@ If you are a Rust project owner and are looking for contributors, please submit
...
 
 An important step for RFC implementation is for people to experiment with the
 implementation and give feedback, especially before stabilization.  The following
-RFCs are at point where user testing is needed before moving forward:
+RFCs would benefit from user testing before moving forward:
 
-<!-- Pre-Stabilization RFCs go here -->
+* [RFC: Deduplicate Cargo workspace information](https://github.com/rust-lang/rfcs/pull/2906)
+    - [Tracking Issue](https://github.com/rust-lang/cargo/issues/8415)
+    - [Testing steps](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/unstable.md#testing-notes)
```
Muscraft, line 83: What're your thoughts on this structure? I know this is what I had but I was worried it was too different of a format from other things around it. 

As well should we use tracking issues or summary posts (where there are ones). Summary posts may give a more in-depth look than just the tracking issue i.e. [Tracking Issue](https://github.com/rust-lang/cargo/issues/8415) vs [Sumamry](https://github.com/rust-lang/cargo/issues/8415#issuecomment-1112618913)

draft/2022-05-18-this-week-in-rust.md
```diff
@@ -72,15 +72,43 @@ If you are a Rust project owner and are looking for contributors, please submit
...
+* [RFC: Packages as (optional) namespaces](https://github.com/rust-lang/rfcs/pull/3243)
```
Muscraft, line 84: Should testing notes always be included in this section? Giving only the RFC without how to test the changes might be confusing

U007D: Maybe?

I think we should gather a couple of weeks of data to determine whether testing notes should be a blocking condition to appearing in TWiR.  As I mentioned above, I'm guessing we're going to see variability, so if you have a strong preference one way or the other, let me know and I'll include that in the thinking that is all but certain to occur over the next couple of weeks.

draft/2022-05-18-this-week-in-rust.md
Muscraft: That's perfect! I am updating some of the cargo docs to include the updates to TWiR. This comment as well as a few others will be helpful as I write those

draft/2022-05-18-this-week-in-rust.md
Muscraft: I was mostly wondering as I am writing up docs for cargo and want to know if we should *require* testing notes. I will go with that we do need testing notes and can update the docs if they aren't required.

U007D: I see.  How about making the recommendation to include testing notes/documentation or other instructions a "strong recommendation" (or some similar verbiage), since we're not blocking inclusion on having the notes (at least not at this point)?

Muscraft: I can change it to that but cargo might be one case where testing notes should be required regardless of TWiR requiring them

nellshamrell: 👍

nellshamrell (approved):
Ty!

PR #3261, opened 2022-05-17
Add scoped threads to Call for Testing

m-ou-se:


U007D:
Hi, @m-ou-se.  Can you point me to anything outlining the kind(s) of testing you would like to see?  I will include this in TWiR-443.

m-ou-se:
Nothing specific I think. The [`std::thread::scope`](https://doc.rust-lang.org/nightly/std/thread/fn.scope.html) function is available in nightly, and it'd be great if folks could use/test it in whatever way they like.

It might be good to include a link the docs: https://doc.rust-lang.org/nightly/std/thread/fn.scope.html (But the tracking issue also links that. And back.)

U007D:
Closing this PR as I have incorporated the data from this PR and from the comment thread into [TWiR #443](https://github.com/rust-lang/this-week-in-rust/pull/3260).

Going forward I would like to create a `call-for-testing` label that authors such as yourself can use.  I will query for these each week to pick up all the relevant items.

PR #3265, opened 2022-05-17
inspect_links: detect truncated link titles

### LINKED ISSUE #3244: Detect truncated link titles
Discord truncates long page titles. If a page is titled **Very very very long thing** then Discord might abbreviate it as **Very very very lo...** (I'm exaggerating; I haven't actually checked the character limit.)

This gets picked up by the bot that @extrawurst wrote, and because it's at the end of a long title, it's easy for humans to miss.

I fixed one by hand in this week's community PR; a few mistakes that weren't detected are shown in #3243.

It would be nice to have either the bot detect this, or have our CI flag it as probably-wrong.

ericseppanen:
Links collected from Discord will sometimes be truncated to exactly 70 characters including a "..." suffix. If the script sees a title like this, raise a warnings as it's probably truncated and should be corrected.

If we're unlucky enough to trigger this warning by mistake, here are some workarounds:
- Make any change to the title so that it's not exactly 70 characters
  (e.g. add an extra space between words)
- Replace the "..." with unicode "…"

Fixes #3244.

### DIFF
--- a/tools/inspect_links.py
+++ b/tools/inspect_links.py
@@ -65,6 +65,24 @@ def is_strict_title(title):
     return title.lower() in STRICT_TITLES
 
 
+def check_truncated_title(tag):
+    """ Flag any links that have a probably-truncated title.
+
+    Links collected from Discord may be truncated to a length of exactly
+    70 characters, including a "..." suffix.
+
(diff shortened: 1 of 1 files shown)


nellshamrell (approved):
Looks great! Ty!

PR #3270, opened 2022-05-18
2022-05-18: Add Evil Twin Attacks in practice (with Rust and a Raspberry Pi)

skerkour:
✌️


ericseppanen:
I'm a bit confused by the article title, as I don't see any Rust content in this post.

nellshamrell:
Hello @skerkour - I have LOVED your Rust blog posts. While I realize this is an excerpt from "Black Hat Rust", there unfortunately is not enough Rust content in the post for us to include it in TWiR. Please keep submitting your Rust content, though!

PR #3271, opened 2022-05-18
Walkthrough "Rust Ergonomics: Default and From"

pop:
Adds simple walkthrough for using the Default and From traits in Rust.

Let me know if there's any changes I need to make or if we need to wait until next week's TWIR to get it added!


ericseppanen (approved):
Thanks!

PR #3283, opened 2022-05-24
Add perf for this week

Mark-Simulacrum:
Omitted the summary table this week per feedback; we're still iterating on the best way to show that data (or whether to show it).


nellshamrell (approved):
Ty!

PR #3290, opened 2022-05-25
Add an observation article

wyhyhyhyh:
Adding a recently posted blog on why Singularity Data moved from C++ to rust for its development of a streaming database product.


ericseppanen (approved):
Thanks!

PR #3292, opened 2022-05-25
Added BonsaiDb performance update blog

ecton:
As always, thank you for putting This Week in Rust together!


ericseppanen (approved):
Thanks!

Moving to Observations/Thoughts since this discussion is quite useful beyond BonsaiDb.

PR #3301, opened 2022-05-28
github actions builds

acim:
I believe this is an interesting blog.

https://ectobit.com/blog/speed-up-github-actions-rust-pipelines/


ericseppanen (approved):
Thanks!

PR #3307, opened 2022-05-31
Add mirrord 2.0 update

eyalb181:
We rewrote mirrord in Rust, thought it might be interesting!


ericseppanen (commented):
While this sounds interesting, I don't see any Rust content in this post.

nellshamrell:
Hello @eyalb181 - this looks like a cool project, but we need more Rust content to include this in TWiR. I encourage you to write a post about the process of rewriting it in Rust, why you chose Rust, and the lessons you learned along the way. We would love to feature something like that.

eyalb181:
Will do, thanks!

PR #3314, opened 2022-06-01
Added BonsaiDb optimization update

ecton:
I'm not sure if this post should be included in the newsletter. In the end I decided to push the link and let the newsletter editors decide. 

It's focused on an optimization I made, and while there are links to pull requests with Rust code, it's not really a Rust post, and there are no actionable items like a new release. Yet, it's an update to a post from last week's newsletter.

I'll be happy whether this link is included or not.

Thank you as always for organizing the newsletter!


nellshamrell:
There's not quite enough Rust content to include this in TWiR, but thank you for submitting this to us :)

PR #3317, opened 2022-06-01
community updates for 2022-06-01

ericseppanen:


ericseppanen:
About the "RustConf Presentation Schedule" link:

@cdmistman suggested that "Events" might be a better place to put it, I agree that sounds good, but it's outside the date range, and it would be disappointing to bury it deep in the event list, so I'm not sure quite what to do.

@nellshamrell Your call...

nellshamrell:
@ericseppanen Miscellaneous makes the most sense until we're closer to the event. Ty for this work! I really appreciate what you all do every single week!

PR #3320, opened 2022-06-03
Add Async I/O in Depth Series Video (Part 1)

nyxtom:
I'm currently in the midst of covering async i/o in rust. Progressively I build an HTTP Web Server and discuss the use of channels, thread pools, a radix tree router, boxed traits, and more. The next video in the series I plan to cover extending this to support a custom async runtime as I cover futures, event pools and using epoll/kqueue/events.


ericseppanen (approved):
Thanks!

FYI, we prefer the canonical link (www.youtube.com/...) to skip the redirect. I'll fix it up this time.

PR #3322, opened 2022-06-03
Build Rust container images faster using layer caching

acim:
This is basically an addition from the last issue article about speeding up GitHub Actions.


nellshamrell:
Hi @acim! Thank you for this pull request. Unfortunately, even though Rust is mentioned, there isn't quite enough Rust material to include it in TWiR. Would you be willing to expand this perhaps to provide a tutorial on creating a piece of Rust software, then using this to build it?

Issue #2781, opened 2022-01-08
Empty pull request list item in issue 422


kraai:
The tenth pull request list item in [issue 422](https://this-week-in-rust.org/blog/2021/12/22/this-week-in-rust-422/) is empty.  The original text is

```
* [add `[T]::as_simd(_mut)`](https://github.com/rust-lang/rust/pull/91479)
```

I think that the Markdown formatter interprets the item body as a reference link definition for <code>add `T</code>.

ericseppanen:
This formatting glitch is really weird. I tried to figure out what's happening:
- It's the right-bracket `]` that is the problem
- but only when it's followed by a colon `]:`

I can't find any explanation of why this doesn't work; the python `commonmark` module handles this correctly, so I can only conclude that it's a bug in the `Markdown` library.

ericseppanen:
As a really dirty hack, if you put a unicode zero-width space character in between the `]` and `:`, then it no longer triggers the bug.

ericseppanen:
Here's a slightly less hacky way to work around the Markdown bug: just add something else to the list item. For example, adding an html space `&nbsp;` at the end of the affected line will stop the bug from happening.

ericseppanen:
Fixed by #2784.

PR #3336, opened 2022-06-08
C/QotW + notable changes

llogiq:


llogiq:
No idea why the markdown check fails. Perhaps someone can help?

ericseppanen:
> No idea why the markdown check fails. Perhaps someone can help?

I will have a look.

ericseppanen:
The error is because of this line:
```
* [implement `[OsStr]::join`](https://github.com/rust-lang/rust/pull/96881)
```

That should be valid markdown, except that Python-Markdown <= 3.3.6 doesn't parse it correctly, which results in an empty list-item in the output, instead of the link we asked for.

This was seen previously as #2781. Since it would silently show up broken on the site, I added a CI check for empty list-items in the output.

I see that Python-Markdown 3.3.7 has been released (which *I think* should have the fix), but until we get that integrated into the site deployment workflow, my workaround is to add a useless `&nbsp` to the offending line, i.e.
```
* [implement `[OsStr]::join`](https://github.com/rust-lang/rust/pull/96881) &nbsp;
```

nellshamrell:
Doing the workaround for now, but will take a look at upgrading (ty for the pull request!) later this week

PR #3340, opened 2022-06-09
Add this week in Databend 45

PsiACE:
Ty!

---

### Databend Supports Async Insert Mode

When thousands of clients concurrently insert a small batch of data, each insert will be executed as follows:

`Parser` -> `Planner` -> `Interpreter` -> `Pipeline`

It's inefficient because of I/O depth and cache locality.

To solve the problem, we want to buffer small inserts into batches in server which sacrifices tiny latency for better insert throughput, smaller block count and larger `DataBlock` in storage.

After doing this, inserts into the same table will be parsed and planned individual. The insert data will be convert to `DataBlock` and buffered. When some conditions are triggered, the buffered `DataBlock` will be interpreted once which also is beneficial for pipelines.

**Learn more:**

- [RFC | Async Insert Mode](https://databend.rs/doc/contributing/rfcs/async-insert)
- [PR 5567 | Support async insert mode to imporve throughput](https://github.com/datafuselabs/databend/pull/5567/)

### DIFF
--- a/draft/2022-06-08-this-week-in-rust.md
+++ b/draft/2022-06-08-this-week-in-rust.md
@@ -36,6 +36,7 @@ and just ask the editors to select the category.
 ### Project/Tooling Updates
 
 * [This week in Databend #44: A Modern Cloud Data Warehouse for Everyone](https://weekly.databend.rs/2022-06-01-databend-weekly/)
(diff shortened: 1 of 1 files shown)


nellshamrell (approved):
Ty!

PR #3342, opened 2022-06-09
Add Qdrant - v0.8.0 update of the Qdrant engine was released

kacperlukawski:
Qdrant is a vector database designed to support neural search with extended filtering capabilities. We've just released version 0.8.0 which experimentally supports distributed mode.


nellshamrell:
Hello @kacperlukawski! Ty for submitting this! Unfortunately, there's not enough Rust content to include this in This Week in Rust, even though the project may be written in Rust. Would you be willing to write something up with examples of Rust code using this database or about decisions you made when implementing them in Rust? Or what you learned about Rust in the process of making this?

PR #3343, opened 2022-06-09
add tidy your rust imports article to walkthroughs

JustinWoodring:


ericseppanen:
Thanks for the link! I have a few questions about the article, if you don't mind.

Most projects import symbols with the more compact syntax, e.g.
```
use crate::models::clubs_md::{Club, NewClub, ClubDetails};
```
If you're bothered by the verbosity of Rust imports it seems unusual to choose the most verbose syntax for this article. And I think this distracts from the point you're making.

I also wonder if `use self::super::prelude;` is correct.  Should this be `use super::prelude::*;` or `use crate::prelude::*;` ?

JustinWoodring:
Thanks for the feedback! I addressed the concerns you mentioned:
1. You were right, self::super::prelude without the ::* would actually not work and even still it looks bad compared to just super::prelude::*;
2. I cleaned up the use example to use the more compact syntax.
3. I addressed using both super::prelude::*; or crate::prelude::*; depending on whether a prelude is being used specifically for a submodule or to organize general types across an entire project.

Let me know what you think :)

ericseppanen:
Looks good to me!

ericseppanen (approved):

PR #3347, opened 2022-06-12
Add Async I/O in Depth Series (Part 2)

nyxtom:
This video is the second part in a series of videos on async i/o in Rust. After last week's implementation of an HTTP Server with channels, thread pools, radix trees, process forking.. etc - this week's video will cover non-blocking I/O. I cover things like epoll, poll, non-blocking TCP, state machines, event loops and the reactor pattern. This video builds up to next week's video on async runtimes.


ericseppanen (approved):
Nice video!

PR #3359, opened 2022-06-15
C/QotW + notable changes

llogiq:


nellshamrell (approved):
Ty! 

And I'm so happy to see a quote from the Mobilization plan included, I contributed to that report :) (though I did not write that quote)

PR #3361, opened 2022-06-15
2022-06-15: Add 42 Companies using Rust in production (2021)

skerkour:
✌️


ericseppanen:
As this is more than a year old, I'm not sure it's on-topic for This **Week** in Rust :)

I'll let @nellshamrell decide...

nellshamrell:
I agree with @ericseppanen that this is too old to feature TWiR. @skerkour you write such good things - I'd love to feature new content from you!

PR #3373, opened 2022-06-17
Add FCP for RFC 3245

tmandry:
It looks like this one got missed from the last issue.

FCP will end a couple days after the next issue goes out. Should it also be added to that one so email readers see it?


nellshamrell:
Apologies, I missed this PR. It has been added to the issue that will go out tomorrow :)

PR #3388, opened 2022-06-22
Blog post on Statime - Precision Time Protocol in Rust

erikjee:
Blog post that announces the open-source implementation of (parts of) the Precision Time Protocol, PTP, that we have been working on.
Thanks!

### DIFF
--- a/draft/2022-06-22-this-week-in-rust.md
+++ b/draft/2022-06-22-this-week-in-rust.md
@@ -50,6 +50,8 @@ and just ask the editors to select the category.
 
 ### Miscellaneous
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
This looks like an awesome project, but there's not enough Rust content for me to include this post in TWiR. However, if you'd like to write up a post on what the experience of writing this in Rust was and what you learned in the process (as well as how someone might use this library in their own code) I would love to feature something like that!

erikjee:
Sorry, you are completely right about that!
I just thought "this is a cool Rust project", but didn't realize the article doesn't go into implementation specifics.

We'll give it a retry when we write a new one. And we'll try to include your "what you learned in the process" suggestion, I really like that idea.

PR #3402, opened 2022-06-26
Add release-plz article

MarcoIeni:
I hope the section is right.
As always, thanks for this awesome newsletter!


nellshamrell (approved):
Ty!

PR #3405, opened 2022-06-27
Add rust malaysia meetup july

pickfire:


andrewpollack:
LGTM -- Thank you for contributing this event to the newsletter!

pickfire:
I have been sending pull requests for malaysia event for quite some time, since we do the event late, maybe next time we should start reserving the place for event.

PR #3406, opened 2022-06-27
Add slint weekly report -- from Embedded World

hunger:


nellshamrell:
Hello! Normally I am happy to include these updates, however I don't think there's quite enough Rust specific content in this update (as it is largely about Embedded World 2022 - which is cool!). I'm not going to merge this one, but I would love to see more Rust content in next week's update!

hunger:
Yeah, we were thinking about whether tonskip this week, since we all were at EW and did not really do too much coding.

I fully understand your reasoning, now that I read it again. It did feel very rusty at the show, as we talked a lot about rust and with rust people, but I did not really write that down.

Thanks for keeping the relevance up on TWiR, even though it cut out my stuff this time:-(

PR #3407, opened 2022-06-27
Add post on `tracing-mutex`

bertptrs:
This blog post describes the algorithm behind the `tracing-mutex` crate and has some background about on why it works to prevent deadlocks in mutex-heavy code.

Disclaimer: this is my crate and blog, and I'm not 100% sure if it's appropriate. The algorithm is not strictly rust, while the safety it provides is similar to the other safety Rust provides and there is some discussion on the implementation, the focus of the article is the algorithm and not the crate.


nellshamrell (approved):
This is great! Ty!

PR #3415, opened 2022-06-28
C/QotW + notable changes

llogiq:
This time, the QotW has an image (funny, I used that in my RustFest Zurich talk to make the same joke in rhyme form with a ukulele solo...good times). We may want to host that somewhere else to avoid unduly burdening the original source.


nellshamrell:
Our pull request checks don't seem to like that image embedded. Going to alter this slightly to link to the image so, if people wish to, they can click on it.

Issue #3412, opened 2022-06-28
Changes to TWiR's Job Listings


nellshamrell:
Hello Rustacean community!

When This Week in Rust started, the Rust language and ecosystem were very young. Jobs which featured working in Rust were few and far between. At the time, it made sense to include them in the newsletter so that the then smallish group of Rust enthusiasts would know where they might be paid to use their Rust skills.

As we have seen, Rust usage has exploded over the past few years - which is a good thing for the language and the community at large. There are many more Rust jobs available - more than we could feature in the newsletter week to week. Additionally, as more Rust jobs have become available, our community has requested that TWiR do more to vet the jobs it lists. Unfortunately, this is not something the TWiR editors have the time or energy to do. All of us work on This Week in Rust on our own time outside of our day jobs. 

To keep TWiR sustainable for its editors (and avoid burnout), we have decided to stop listing specific job postings in the newsletter. Instead, we will link to the current "Who's Hiring" thread in the Rust subreddit (the most recent of which can be found [here](https://www.reddit.com/r/rust/comments/utav5l/official_rrust_whos_hiring_thread_for_jobseekers/)). The Rust community does a good job of setting expectations for jobs that are posted to the thread and the /r/rust moderators do an excellent job of moderating posts which do not follow those expectations. 

We would like to thank all of the companies which have submitted job listings to This Week in Rust over the years - I have no doubt they helped Rustaceans find jobs where Rust usage was welcomed and encouraged. I encourage all employers looking for Rustaceans to submit posts to the "Who's Hiring" thread in /r/rust.

Thank you!

Nell Shamrell-Harrington
Lead Editor, This Week in Rust

PR #3417, opened 2022-06-29
Add WalletConnect job

agarella:


nellshamrell:
Hello! Thank you for submitting this job posting. We are making some changes to how we handle job listings in This Week in Rust, full details can be found in https://github.com/rust-lang/this-week-in-rust/issues/3412.

PR #3419, opened 2022-06-29
community updates for 2022-06-29

ericseppanen:


ericseppanen:
I'm torn about whether to post the pre-release testing announcements (Inside Rust blog), because the timing is bad: by the time the TWiR emails go out, the actual release will have happened.

nellshamrell:
I think it's ok - it alerts people to look for them next time.

nellshamrell (approved):
Ty so much!

PR #3420, opened 2022-06-29
DRAFT_TEMPLATE: remove the jobs section

ericseppanen:
Leave a hidden note for contributors, with a link to the GitHub issue.

The note could be a bit friendlier, but I'm not sure quite what to add... "So long, and thanks for all the jobs"?


nellshamrell:
Made it *slightly* more friendly

PR #3425, opened 2022-06-30
Update 2022-07-06-this-week-in-rust.md

mitghi:
Add link to [quartz_sched.rs](https://crates.io/crates/quartz_sched)


nellshamrell:
Hi @mitghi! We don't generally include links to crates.io or repos without there being more content in the README (and we strongly prefer blog posts). Do you mind writing up a blog post on the process of creating this project, what you learned about Rust in the process, and how to use the project? I would love to feature something like that.

PR #3440, opened 2022-07-05
Events updates for 450

andrewpollack:
A couple events are hybrid (in-person and livestreamed virtually). For those, I've opted to duplicate across its continent and the "Virtual" sections


mariannegoldin (approved):
Looks great! I think the duplication of live streamed events in the in-person section is a helpful aid for the reader. Approving.

Issue #27, opened 2015-01-25
Link to Alternative Breaking Changes Listing


killercup:
Recently, I [rewrote](https://github.com/killercup/bitrust) the code for [bitrust](http://bitrust.octarineparrot.com/) in Rust and added a nice HTML frontend with filtering and markdown for commit messages. I hope that it offers a good overview of what changed in Rust.

You can [find it here](http://killercup.github.io/bitrust/).

Maybe you'd like to link to this in the next post, in addition to bitrust.

PR #3482, opened 2022-07-22
Fix date for Hack Night #27 in Copenhagen

troelsarvin:
See https://www.meetup.com/copenhagen-rust-meetup-group/events/287057762/


andrewpollack:
Thank you!

PR #3487, opened 2022-07-26
Add Tokyo Rust Game Hack 2022 edition: Bombercrab

strohel:
Is it okay to add an entry to both Virtual and Asia sections for a hybrid event? Each section entry links to slightly different, dedicated page.

CC @PabloMansanet.


andrewpollack:
Absolutely okay to add to both sections! Thank you for your contribution 😸

PR #3503, opened 2022-07-30
Add Rust Melbourne meetup for August 2022

alfiedotwtf:


andrewpollack:
Looks like a fun event 😸 thank you for the PR!

PR #3509, opened 2022-08-01
Add RRust blog post

Erk-:
I have added it under `Research` since I find it fits there best, but it may fit better in observation/thoughts


nellshamrell (approved):
Thank you!

PR #3511, opened 2022-08-02
C/QotW + notable changes

llogiq:


draft/2022-08-03-this-week-in-rust.md
```diff
@@ -45,7 +45,9 @@ and just ask the editors to select the category.
 
 ## Crate of the Week
 
-<!-- COTW goes here -->
+This week's crate is [lending-iterator](https://lib.rs/crates/lending-iterator), a type similar to `std::iter::Iterator`, but with some type trickery that allows it to `.windows_mut(_)` safely.
```
Globidev, line 48: Small nit but it probably should be
> a trait similar to ...

Rather than

> a type similar to ...

Ben-Lichtman: `but with some type trickery that allows it to .windows_mut(_) safely` Doesn't really do this crate justice.

maybe something more like "frees you from the shackles of Iterator" or "finally allows you to do this thing that everyone wants to do all the time but can't"

nellshamrell (approved):
Ty!

PR #3519, opened 2022-08-04
Add this week in Databend 53

PsiACE:
ty!

---

### Call for Migrating Functions to the New Expression Framework

If you are interested in typed type system, or maybe you'd like to try your hand at a database project, take a look at how Databend does it.

We are now trying to migrate some old functions to the new expression framework, would you like to try it out?

- [Migrate String functions to new expression framework #6766](https://github.com/datafuselabs/databend/issues/6766)
- [Migrate control-flow functions to new expression framework #6833 ](https://github.com/datafuselabs/databend/issues/6833)

**Background**

Recently Databend is working on a new expression framework that will bring some interesting features.

- Type checking.
- Type-safe downcast. 
- Enum-dispatched columns.
- Generic types.

**How To**

Legacy functions are settle in `common/functions/src/scalars`. The task is to migrate all of them to `common/functions-v2/src/scalars/`. 

Usually you can reuse the logic of the previous implementation, it just needs some rewriting to make it fit the new way.

Similarly, the legacy tests in `common/functions/tests/it/scalars/` should also be migrated to `common/functions-v2/tests/it/scalars/`. 

The new tests will be written using `goldenfile`, so you can easily generate test cases without a lot of painful handwriting. 

**Example**

A unary function `OCTET_LENGTH` can be defined using 6 lines in`common/functions-v2/src/scalars/strings.rs`.

`OCTET_LENGTH` will return the length of a string in bytes.

```rust
registry.register_1_arg::<StringType, NumberType<u64>, _, _>(
    "octet_length",
    FunctionProperty::default(),
    |_| None,
    |val| val.len() as u64,
);
```

`LENGTH` is a synonym for `OCTET_LENGTH`. 

We can easily define function aliases with one line.

```rust
registry.register_aliases("octet_length", &["length"]);
```

Next, let's write some tests to make sure it works correctly. 

Edit `common/functions-v2/tests/it/scalars/string.rs`.

```rust
fn test_octet_length(file: &mut impl Write) {
    run_ast(file, "octet_length('latin')", &[]);
    run_ast(file, "octet_length(NULL)", &[]);
    run_ast(file, "length(a)", &[(
        "a",
        DataType::String,
        build_string_column(&["latin", "кириллица", "кириллица and latin"]),
    )]);
}
```

Register it in the `test_string` function, 

```rust
#[test]
fn test_string() {
    let mut mint = Mint::new("tests/it/scalars/testdata");
    let file = &mut mint.new_goldenfile("string.txt").unwrap();

    ...
    test_octet_length(file);
    ...
}
```

Next, let's try to generate these test cases from the command line.

```bash
REGENERATE_GOLDENFILES=1 cargo test -p common-functions-v2 --test it
```

Well done, we did it.

**Learn More**

- [RFC: Formal Type System](https://github.com/datafuselabs/databend/discussions/5438)
- [Tracking issue for new expression framework](https://github.com/datafuselabs/databend/issues/6547)

### DIFF
--- a/draft/2022-08-03-this-week-in-rust.md
+++ b/draft/2022-08-03-this-week-in-rust.md
@@ -35,6 +35,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
+* [This week in Databend #53: A Modern Cloud Data Warehouse for Everyone](https://weekly.databend.rs/2022-08-03-databend-weekly/)
+
 ### Observations/Thoughts
 
 ### Rust Walkthroughs


nellshamrell (approved):
Ty!

PR #3527, opened 2022-08-09
Adding publish automation

andrewpollack:
As from README, this Dockerfile takes care of installation of `pelican`, `sass`, `juice`, and other requirements.

Example flow (from `publishing` directory):
1. `make build && make generate-website && make host-content`
2. `make build && make generate-email && make host-content`, change "BLOG_DOWNLOAD" variable, and in separate terminal run `make optimize-email`

Many improvements to do in future iterations...
1. Substitute "BLOG_DOWNLOAD" variable for read-in variable (or better automation)
2. Move `plugins/`, `themes/`, `pelicanconf.py`, `requirements.txt` into `publishing/`


publishing/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:18.04
```
ericseppanen, line 1: That's pretty old, and has less than a year of support left. Something newer would be nice?

nellshamrell:
Hi @andrewpollack! This is awesome! Will merge this. I'm still considering this experimental for the time being, but I think it is an awesome alternative to the currently highly dependent on environment set up we have right now.

Excellent work!

nellshamrell (approved):

PR #454, opened 2017-06-11
Add What's coming up in imag (26)

matthiasbeyer:
Update from the imag project.

PR #3535, opened 2022-08-10
community updates for #454

ericseppanen:
There's an open question about whether "Bevy Jam # 2" belongs in community links, events, or both. I included it in the Miscellaneous section for the moment...


andrewpollack:
RE: Bevy -- this will be good as a part of community links

Thank you!

PR #3546, opened 2022-08-12
Add Ledger fault injection blogpost

erdnaxe:
Maybe alternate title that would be clearer: "Integrating fault injection evaluation in Rust development workflows"

This blogpost shows an integration between some no_std embedded Rust crates and a dynamic-analysis framework to detect single-fault vulnerabilities. It shows some example of contermesures in the case of an ARM-Thumb target.

Best,


ericseppanen (approved):
Interesting article; thanks!

PR #3549, opened 2022-08-14
add erased serde article

jhartzell42:
Relatively beginner-friendly walk-through with somewhat practical examples on why and how to use dynamic vs static polymorphism, type erasure, the erased-serde crate, and the newtype idiom.


ericseppanen (approved):
Thanks!

PR #3552, opened 2022-08-16
Prettier code formatter for Rust

jinxdash:
Prettier code formatter for Rust - https://github.com/jinxdash/prettier-plugin-rust

I'm aware of the strong recommendation against linking repos directly. I believe the readme is good enough.


ericseppanen:
Thanks for the contribution, but as you noticed, we don't usually publish just links to project's github repositories. Perhaps you would consider writing up a blog post discussing the project instead?

PR #3557, opened 2022-08-16
C/QotW + notable changes

llogiq:


mariannegoldin (approved):
Thanks so much, approving.

ericseppanen:
Hi @llogiq , I was noticing that there are two bits of markdown in this PR that don't render correctly:

```
 * [make `\[rust\] use-lld=true` work on windows](https://github.com/rust-lang/rust/pull/100464)
```
and
```
 * [rust-analyzer: infer byte string pattern as `&\[u8\]` when matched against slices](https://github.com/rust-lang/rust-analyzer/pull/12992)
 ```

These backslashes aren't necessary, and they actually show up in [the rendered issue](https://this-week-in-rust.org/blog/2022/08/17/this-week-in-rust-456/).

Is there a script that generates this list? I'm willing to investigate but I don't really know where to look.

llogiq:
No, I added them by hand because my markdown parser would otherwise balk. But now I know it I'll avoid that in the future.

PR #3559, opened 2022-08-17
tools: improve filename regex in inspect_links

ericseppanen:
When merging, I sometimes leave `*.orig` files behind, and inspect_links starts yelling about a lot of duplicate links. A little tweak to the regex fixes that.


mariannegoldin (approved):
Thank you, approved.

PR #3563, opened 2022-08-17
Add This Week in Fluvio 42

tjtelan:
Thanks!


mariannegoldin (approved):
Thank you, approving. In the future, we would encourage including more Rust content in your updates, about how to use the project.

PR #3565, opened 2022-08-18
Add this week in Databend 55

PsiACE:
ty!

---

Let's take a look at what's new at Datafuse Labs & Databend each week.

**common table expressions**

Databend supports common table expressions (CTEs) and allows you to use a WITH clause to define one or multiple named temporary result sets that are used by the query that follows. The "temporary" means that the result sets will be not permanently stored anywhere in the database schema. They act as temporary views that are only available to the query that follows.

Learn more: <https://databend.rs/doc/reference/sql/query-syntax/dml-with>

**crates**

- [openraft](https://github.com/datafuselabs/openraft) has released v0.7.0, which includes a number of refactorings and reliability improvements, and adds examples of rocksdb and sled.
- [opendal](https://github.com/datafuselabs/opendal) has just added support for Google Cloud Storage and released 0.13.0. Recent improvements include a new `Builder` with sync support and some useful layers.
- The v0.2.0 release of [opensrv](https://github.com/datafuselabs/opensrv) supports streaming writes to mysql result sets and simplifies metadata for clickhouse.

### DIFF
--- a/draft/2022-08-17-this-week-in-rust.md
+++ b/draft/2022-08-17-this-week-in-rust.md
@@ -35,6 +35,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
+* [This week in Databend #55: A Modern Cloud Data Warehouse for Everyone](https://weekly.databend.rs/2022-08-17-databend-weekly/)
+
 ### Observations/Thoughts
 
 ### Rust Walkthroughs


mariannegoldin (approved):
Thank you, looks like a lot of great work from many awesome contributors. Approving.

PR #3574, opened 2022-08-23
Update 2022-08-24-this-week-in-rust.md

applied-math-coding:
An interesting application of Rust within a web-based single-player logic game.


ericseppanen:
Unfortunately, there isn't enough Rust content in this article to be included in TWiR. If you publish a blog post that shows how the code works, and discusses some of the design decisions involved, that's something we'd like to hear about!

PR #3576, opened 2022-08-23
Add slint project updates

hunger:
We had little news while everybody was on summer break, so we did not send in anything for the last few weeks. But we are back to hacking again.


mariannegoldin (approved):
Looks good, we enjoyed the note about `const` usage. Approving.

PR #3580, opened 2022-08-24
fix a bunch of old backslash formatting issues

ericseppanen:
Most of these are of the form `` `\[T]\` -> `[T]` `` as brackets don't need to be escaped inside backticks.

One example is a mix of brackets outside of backticks-- markdown seems to handle this very inconsistently, so just add backticks.

Per discussion on #3557, we shouldn't expect this to happen much in the future.


mariannegoldin (approved):
The updates look great, thanks for taking the time to make them. Approved.

PR #3593, opened 2022-08-28
Add proposed rust walkthrough

rsdlt:
Add "Ruxel - Building a Ray Tracer with Rust Part 1" in Rust Walkthroughs


rsdlt:
Thank you Eric! Will follow the style link for future submissions. Best!

PR #3602, opened 2022-08-30
Rust and C++ Cardiff meetup + recent videos

crowlandsimms:
Added the next meetup event and two recent youtube video uploads.

The videos were uploaded two weeks ago rather than in the last week, but they haven't been posted on This Week in Rust before, so I am hoping it's still okay to add them in case folks are interested in them! The multithreaded one in particular we ran into some weirdness with locking so am hoping someone who watches it might have some ideas of the cause. No worries if it's too late - we'll be more diligent with future uploads.

### DIFF
--- a/draft/2022-08-31-this-week-in-rust.md
+++ b/draft/2022-08-31-this-week-in-rust.md
@@ -40,6 +40,9 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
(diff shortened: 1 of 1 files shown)


mariannegoldin (approved):
We appreciate these submissions, and thank you for adding the calendar item as well. Approving.

PR #3606, opened 2022-08-31
2022-08-31: Add Black Hat Rust "Back2School" Discount

skerkour:
✌️


mariannegoldin (commented):
Thank you for the submission. As we have already featured this book and the related course in issues 442 + 451, we will not be approving this PR. If there is updated content for this book and/or course, please feel free to submit another PR.

PR #3628, opened 2022-09-07
Update 2022-09-07-this-week-in-rust.md

rsdlt:
Add proposed blog post "How to use Rust Traits, Generics and Bounds


ericseppanen (approved):
Nice article, thanks!

PR #3636, opened 2022-09-08
gnu `ld` discards code

U007D:
I hope this is the right section to post this article; please move if appropriate.


U007D:
Approved.  (Cannot self-approve).

PR #3664, opened 2022-09-19
Add RustFi Hackathon to upcoming events in Europe

agarella:


andrewpollack:
Thank you for the PR!

PR #3698, opened 2022-09-29
add cargo-careful

RalfJung:


RalfJung:
Hm I guess now there's little point in still adding this, since who's going to still look at an old TWiR?

Would it be possible to add this next week, since it was missed this week?

andrewpollack:
Absolutely @RalfJung ! Please shift it to rebase off latest main, and add to this week's draft: draft/2022-10-05-this-week-in-rust.md

PR #3700, opened 2022-10-03
2022-10-05: Add Learning Rust: The derive attribute

skerkour:
✌️


andrewpollack:
Thank you for the submission. Unfortunately, this walkthrough is lighter on the details behind the covered subject. Please see https://github.com/rust-lang/this-week-in-rust#what-do-we-look-for-when-considering-whether-to-include-something-in-this-week-in-rust

PR #3702, opened 2022-10-03
2022-10-05: Add "How to call a C function from Rust"

vanjacosic:
It's a pretty simple tutorial, but maybe someone will find it useful 😊


andrewpollack:
Thank you for the submission, it looks great as a tutorial! Unfortunately, for Rust Walkthroughs, we do not link solely to a GitHub repo. If you are able to convert this to a blog-post, we'd be happy to include in this or next week's issue!

For more info on what is included in TWiR, see: https://github.com/rust-lang/this-week-in-rust#what-do-we-look-for-when-considering-whether-to-include-something-in-this-week-in-rust

PR #3707, opened 2022-10-05
2022-10-05—feat: Rust DC event title

scooter-dangle:
Rust DC has a cool talk this month (🎉)


andrewpollack:
Indeed looks like a very cool talk 😎 thank you!

PR #3708, opened 2022-10-05
2022-10-05: Add When to vendor (or not) your dependencies in Rust

skerkour:
✌️


andrewpollack:
Thank you for the submission. Unfortunately, there is not enough Rust-specific content in the post for us to include it in TWiR

PR #3723, opened 2022-10-12
2022-10-12: Add When to vendor (or not) your dependencies in Rust

skerkour:
Hi,

Following [last week's feedback](https://github.com/rust-lang/this-week-in-rust/pull/3708), I've added more details about vendoring Rust dependencies and its specificities (such as very very large `vendor` folders).

✌️


andrewpollack:
Thank you for the follow-up @skerkour! This looks great -- merging now

PR #3724, opened 2022-10-12
2022-10-12: Add update announcement for error-stack

TimDiekmann:
Includes the update announcement for `error-stack` 🙂


andrewpollack:
Thank you!

PR #3729, opened 2022-10-12
added this week's calls for participation…

bennyvasquez:
up to https://users.rust-lang.org/t/twir-call-for-participation/4821/472


andrewpollack:
Awesome, thank you @bennyvasquez !

PR #3733, opened 2022-10-13
Introducing BastionAI

cchudant:
Hello!
This is an article about a new rust project that we announced


mariannegoldin (approved):
This is a great writeup, and we appreciate the time spent introducing the project and presenting design considerations and tradeoffs. Approved.

PR #3737, opened 2022-10-18
Add `Rust on Espressif chips` quarterly update

SergioGasquez:


mariannegoldin (approved):
Thank you for the submission, we appreciate the summaries for each section. 

For future updates, we would also love to see examples/guidance for how to use the added functionality.

Approving.

PR #3731, opened 2022-10-13
submitting a post for Rust walkthrough

anupj:
A post about setting up neovim for Rust dev - https://rusty-ferris.pages.dev/blog/using-nvim-for-rust-development/

PR #3748, opened 2022-10-20
TWiR 465: Add walkthrough link in submission from anupj

mariannegoldin:
Adding a link from @anupj via PR #3731.


mariannegoldin (commented):
Submitting in lieu of PR #3731.

PR #3751, opened 2022-10-20
Added guidance around formatting for the CFP section

bennyvasquez:
I noticed this week while doing the CFP update that the other sections had 'thing goes here' tags. When I took over this section I'd used past issues to identify a pattern for how the section was structured. In this PR I've added guidance to the template for future editions.


mariannegoldin (approved):
This is a helpful example, thanks! Approving.

PR #3756, opened 2022-10-25
Add Zellij release post

imsnif:
Hey there - we would be happy to be included!


mariannegoldin (approved):
Thanks for the update, approving.

PR #3767, opened 2022-10-27
TWiR 467: Add Stockly Event in Paris

mariannegoldin:
Adding this as a one-off (in addition to the bulk events update) to make sure it gets into the next issue.


andrewpollack:
Thank you!

PR #3769, opened 2022-10-27
TWiR-466: Point whos hiring link to the correct one

dswij:
The link for hiring thread points to the question mega thread. This PR updates it to the hiring thread instead.


mariannegoldin (approved):
We appreciate you catching this, will make this update. Approved.

PR #3780, opened 2022-11-01
Add Kani blog about Bolero integration

adpaco-aws:
Adds the Kani blog post on Bolero integration.

(I think "Project/Tooling" updates is the best place to add it, but please let me know if this is not the case. Thanks!)


U007D (approved):
Looks good, thank you!

PR #3800, opened 2022-11-09
Add floating point arcade walkthrough

CrockAgile:
"Floating Point Arcade" is a walkthrough of floating point numbers and random number generation. The code snippets are Rust, and reference the popular `rand` crate.


U007D (approved):
Great post, thank you!

PR #3808, opened 2022-11-15
triage report for this week

pnkfelix:
Once https://github.com/rust-lang/rustc-perf/pull/1496 lands then the link to the full report should work.


bennyvasquez:
Thank you!

PR #3823, opened 2022-11-17
What's new in SeaORM 0.10.x

tyt2y3:
https://www.sea-ql.org/blog/2022-11-10-whats-new-in-0.10.x/


tyt2y3:
My bad, it seems to be included in this-week-in-rust-469 already

PR #3839, opened 2022-11-28
Add blog post on UB-related falsehoods believed by programmers.

obi1kenobi:
It covers falsehoods about undefined behavior I've heard from people over the years. Rust is great here because of the rule that "safe Rust is* UB-free" (*yes, there are some outstanding unsoundness issues, but not that many and they rarely come into play).

Thanks for continuing to run TWIR! Every issue is full of awesome 💯


nellshamrell:
Looks great! Ty!

PR #3844, opened 2022-11-29
Add link to Rust developers survey

bravit:
Hi! 

I consider this survey as being of general interest to the Rust community. Its goal is to learn how exactly Rust and C/C++ ecosystems coexist. The results are might be shared back with the community.


nellshamrell (approved):
Ty!

PR #3854, opened 2022-12-05
Add link to the blog post about Rust macros

bravit:
Hi! Although this [blogpost](https://blog.jetbrains.com/rust/2022/12/05/what-every-rust-developer-should-know-about-macro-support-in-ides/) is coupled with the recent [release of the IntelliJ Rust plugin](https://intellij-rust.github.io/2022/12/05/changelog-184.html), I believe its topic is more general than that, and it's more appropriate for the "Observations/Thoughts" section, not the "Project/Tooling Updates".


nellshamrell (approved):
Ty!

PR #3857, opened 2022-12-06
Draft: Add a Dark Mode

### LINKED ISSUE #2274: Feature request: Dark theme for the website
Thank you for the amazing newsletter 🧡

On the website, it could be nice to have:
- selectable light vs dark mode, with a simple button for picking the active theme, e.g. [dark-mode-toggle](https://github.com/googlechromelabs/dark-mode-toggle) or similar
- possibly autodetect the user-preferred style with `prefers-color-scheme` [css media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) (also supported by above JS snippet and others)
- possibly even support the variety of themes that e.g. `rustdoc` and `mdbook` supports (Light, Rust, Coal, Navy, Ayu)

ndd7xv:
Fixes #2274 (if merged)

Hello! Saw this on TWiR and thought I'd give it a go; I have no experience with frontend work so I'm completely OK with scrapping this PR too.

This sets dark mode for users who have dark color schemes preferred. I tried replicating a similar color scheme to docs.rs' Dark theme because I thought it'd look nice. I'm also down to change the color scheme, but I kind of agree about not liking straight white-on-black, which is discussed in a previous attempt to bring a dark mode (https://github.com/rust-lang/this-week-in-rust/pull/1089#issuecomment-565978373).

Preview:
image
image

I understand that this doesn't give users the option to change between themes. If that prevents a merge, I'll try and spend additional time to implement some tri-state widget as mentioned by @flying-sheep.

---

I also added a tiny commit that's unrelated to dark mode that fixes a typo (developments in Rust community -> developments in THE Rust Community). In addition, I added margins to make it look nicer and aligned with the other components like the input and header text. Without the margins:

image
image

With margins and fixed typo:

image
image

Adding margins is kind of a subjective choice, and  my choice in color scheme is also just my preference, so please let me know what you all think!

### DIFF
--- a/themes/rusted/static/css/_base.scss
+++ b/themes/rusted/static/css/_base.scss
@@ -126,9 +126,9 @@ blockquote {
 pre,
 code {
     font-size: 15px;
-    border: 1px solid $grey-colour-light;
+    border: 1px solid $code-border-color;
     border-radius: 3px;
-    background-color: #eef;
+    background-color: $code-bg-color;
 }
 
 code {
@@ -138,6 +138,7 @@ code {
 pre {
     padding: 8px 12px;
     overflow-x: scroll;
+    color: $pre-color;
 
     > code {
         border: 0;
--- a/themes/rusted/static/css/_base.scss
+++ b/themes/rusted/static/css/_base.scss
@@ -92,7 +92,7 @@ a {
     text-decoration: none;
 
     &:visited {
-        color: darken($brand-color, 15%);
+        color: $brand-color-visited;
     }
 
     &:hover {
--- a/themes/rusted/static/css/_homepage.scss
+++ b/themes/rusted/static/css/_homepage.scss
@@ -8,5 +8,6 @@
 }
 
 .subtext {
-    
+    margin-left: 50px;
+    margin-right: 50px;
 }
--- a/themes/rusted/static/css/_layout.scss
+++ b/themes/rusted/static/css/_layout.scss
@@ -17,9 +17,9 @@
     letter-spacing: -1px;
     margin-bottom: 0;
 
-    &,
+    color: $text-color;
     &:visited {
-        color: $grey-colour-dark;
+        color: $text-color;
     }
 }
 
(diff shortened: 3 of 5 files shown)


nellshamrell:
Ty so much! I will get this reviewed either before or during the weekend!

flying-sheep:
I like it! The only thing that jumps out to me as something I’d change is that the `<code/>` elements are still dark-on-light. I think they should have less contrast with the background and be light-on-dark in dark mode.

ndd7xv:
I modified the code/pre block so it looks like this (screenshot of what https://this-week-in-rust.org/blog/2017/06/20/this-week-in-rust-187/ would look like in dark mode):

Screenshot 2022-12-08 at 17-30-37 This Week in Rust 187 · This Week in Rust

nellshamrell:
That does look nice! I do want there to be an option to switch between dark and light mode, though, as people have different preferences (and sometimes it depends on the light of the room you are in, etc.)

flying-sheep:
If someone has a preference, they can configure it in their OS or their browser.

So the only advantage would be if someone wants dark mode, but is on windows 10 or lower and is forced to use a browser that doesn’t have an addon with a toggle switch. Given the [complexity of the implementation](https://github.com/python/peps/pull/2552), I think adding a toggle should probably go into a separate PR: keep this one focused on style, the next one can be bike shedding about how to best implement a toggle.

holly-hacker:
Is there anything blocking this PR? It seems quite useful :)

nellshamrell:
I think this is good to merge! Ty so much @ndd7xv!

ndd7xv:
Thanks! With this change live, I noticed I hadn't changed some CSS for tables! I'll try and do that tonight since it shouldn't take too long.

I'll also open an issue/request to implement a tri-state toggle or button to change the theme - sorry I didn't get to that, I've been kind of busy and I don't know to much in this domain.

Issue #3669, opened 2022-09-20
Add anchors for headers


scottmcm:
I was trying to make a link to the `## Final Comment Period` section specifically today, and was surprised that there's no `id`/`name` attributes in the page that could be used for that.

Could the generator be updated to add those?

bennyvasquez:
I'm going through and reviewing older issues here, and this is definitely one we see the value of! I've added the 'help wanted' label, and we'll include a reference to this issue in a future edition of TWiR to see if we can get some help with it. Thanks for the submission!

PR #3863, opened 2022-12-06
add header links

ndd7xv:
Fixes https://github.com/rust-lang/this-week-in-rust/issues/3669 if merged

Hi again! Same author of the dark mode draft PR; basically just searched up how to add anchors in pelican and used the [first link](https://github.com/getpelican/pelican/issues/1357#issuecomment-362075129) with [some](https://docs.getpelican.com/en/latest/settings.html?highlight=MARKDOWN#basic-settings) [documentation](https://python-markdown.github.io/extensions/toc/#usage) to get the following result:

image

Again, no experience with UI and my design choices are subjective, but I chose to have a permalink with this specific css to prevent it from being to obtrusive and taking the same colors as other links, while also making it easy to find.

**Let me know if there's anything I can/should change!**

### DIFF
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -48,3 +48,15 @@ scattered about.
 }
 
 PLUGINS = ['webassets']
+
+MARKDOWN = {
+    'extension_configs': {
+        'markdown.extensions.codehilite': {'css_class': 'highlight'},
+        'markdown.extensions.extra': {},
+        'markdown.extensions.meta': {},
+        'markdown.extensions.toc': {
+            'anchorlink': True,
+        },
+    },
+    'output_format': 'html5',
+}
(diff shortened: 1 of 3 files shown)


ndd7xv:
Also note that I disabled in the newsletter because they were relative linking; the link in the README is dead (http://zurb.com/ink/inliner.php) but I tried testing the inline emails by creating an account with app.postdrop.io

nellshamrell:
Ty so much! I will get this reviewed either before or during the weekend!

nellshamrell:
Taking a look now!

nellshamrell:
Hi there! Ty so much for this! Is there any way to make the text of the header the anchor, rather than adding the anchor symbol to the end of the header?

ndd7xv:
Thanks for the review! Made the text the anchor and overrode its color to remain the same as before instead of having it turn blue like the other hyperlinks.

Let me know if there's anything else!

nellshamrell:
This looks FANTASTIC! Ty so much @ndd7xv! Merging!

PR #3873, opened 2022-12-12
1 link: Rust observability post (on LogRocket)

zeeshanlakhani:
These are somewhat "self-motivated", but I wrote a [post for LogRocket](https://blog.logrocket.com/composing-underpinnings-observable-rust-application/) on observability, tracing, and tracing subscriber hooks and such that may be useful.


mariannegoldin (approved):
Interesting topic, thanks for the PR! Approved.

PR #3874, opened 2022-12-12
Videos of talks at Rust meetup in Amsterdam NL

erikjee:
Talks about the Network Time Protocol in Rust, Rustls (TLS in Rust), GATs and packaging for DEB/RPM etc of Rust projects .

Not sure about the category, I'll leave that up to you :)

Thanks!


mariannegoldin (approved):
Thank you, approved

PR #3875, opened 2022-12-12
Blog article on sorting with SIMD in Rust

erikjee:
The article explores if SIMD can be used for sorting, an atypical use-case.


mariannegoldin (approved):
Interesting article, thank you! Approved

PR #3895, opened 2022-12-21
Our year in Rust - a rhyme

erikjee:
Hi, a strange contribution maybe, but definitely "miscellaneous". 

Thanks!


andrewpollack:
Very fun and in the spirit of the season, thank you!

PR #3896, opened 2022-12-21
Docs as Code: Mermaid inline diagrams

frehberg:
Docs as Code: Mermaid inline diagrams - provided by Crate Aquarmarine for Rust  everybody should know


andrewpollack:
Thank you!

PR #3900, opened 2022-12-21
Add Comprehensive Rust 🦀

mgeisler:
Hi all, I'm not sure if this is the right section. Basically, I work on Android and we've published a four day Rust course. We would love to let the community know about it. The course covers Rust in general and is not Android specific.


andrewpollack:
Looks fantastic, thank you!

PR #3924, opened 2023-01-03
Add blog post: "Zero-dependency random number generation in Rust"

orhun:
Hello!

I wrote this blog post about generating random numbers without dependencies in Rust and I think it might be interesting for TWIR readers 🐻

Thank you!


andrewpollack:
Interesting indeed, thank you!

PR #3930, opened 2023-01-04
Add Databend 2022 Recap

PsiACE:
I would love to include Databend 2022 Recap in this week's newsletter.


andrewpollack:
Thank you!

PR #3937, opened 2023-01-06
fix: table rows in dark mode

ndd7xv:
I had accidentally left out colors for the dark theme when it came to tables; my bad!

Before
image

After
image

As always, open to any critiques or changes to the colors.


andrewpollack:
Thank you for spotting and adjusting this!

andrewpollack:
Site is updated to reflect the change

Issue #2569, opened 2021-11-03
Add a search


Mazwak:
Hello

It would be nice to be able to search on articles title.

And see a list of all article, with a date.

Regards

bennyvasquez:
I'm going through and reviewing older issues here, and this is definitely one we see the value of! I've added the 'help wanted' label, and we'll include a reference to this issue in a future edition of TWiR to see if we can get some help with it. Thanks for the submission!

PR #3942, opened 2023-01-09
add search

ndd7xv:
Hello! Created a quick PR of a search for TWiR, though it comes with some caveats:

- This uses the search plugin for Pelican - there don't seem to be many options for search functions in Pelican. There was [tipue_search](https://github.com/pelican-plugins/tipue-search), but the repository says it's deprecated and they recommend to use the plugin I mentioned
- This was to fix https://github.com/rust-lang/this-week-in-rust/issues/2569, but the search isn't perfect. Searching for unique words/phrases generally works (e.g. "Microsoft" for any post with Microsoft, or "444" for TWiR 444), but if there's too general of a phrase (e.g. "This Week in Rust 1", since "This", "Week", "in", "Rust" and "1" appear in many posts) there will be too many results that it won't show up (I think stork just shows the first 20). As far as I can tell, there is no exact matching.
-  The search plugin relies on [Stork Search](https://stork-search.net/) (which the search plugin uses, written in Rust!). As such, and per the [plugin's README](https://github.com/pelican-plugins/search#installation) I have to install it. I'm not 100% sure how TWiR is deployed, but I made changes to the Dockerfile so that it worked when I ran the image locally (with `make clean`, `make build`, `make generate-website` and `make host-content`).
- **To test search (locally and locally in the container), I had to change the `SITEURL` in `pelican.conf` `to http://localhost:8000`.** If I didn't, then locally running Pelican would have its hyperlinks reference https://this-week-in-rust.org, which wouldn't have the files that I added to make search work (like `themes/rusted/static/css/{stork-dark, stork}.css` and `themes/rusted/static/js`). Not sure of a better way/best practice to test my changes, but let me know if there is one.
- Kind of a small note, I just added this on the main page. Didn't think there was a good place to put it nicely on an article page or anywhere else

Given this, I understand if this isn't comprehensive enough to be merged, but I thought I'd make this for posterity's sake. Hopefully exact matching and more comprehensive search functionality is added when Stork Search 2.0 is released (its [roadmap](https://github.com/users/jameslittle230/projects/1/views/1) has 'Advanced query searches', but let me know if there's anything I can tweak or change!

Results:
Screenshot 2023-01-08 at 21-44-09 This Week in Rust


bennyvasquez:
This actually looks great, even if a bit simple. We're still discussing as an editorial team, but I think as long as we add some guidance to the interface about what will and won't work, this would be a good option. Is that something you'd like to add, or would you like help with that?

blueglyph:
Couldn't a magnifying glass fit in the header, in `base.html`, and open the search bar when clicked? It's a common pattern, for ex. on [the book](https://doc.rust-lang.org/book/), [the reference](https://doc.rust-lang.org/reference/index.html) and other Rust-related websites, and that would make it available on each page.

image

I'm not equipped right now to try and modify it and frankly, not familiar enough with stylesheets so at best it would just be a hack. So it's just a suggestion, a search on the front page is already a great addition IMO, thanks for taking the time @ndd7xv. 🙂

ndd7xv:
@bennyvasquez I've been a little bit busy, but I'd still love to be able to help out! I think I can add guidance to the interface; would you be willing to describe what you're imagining it would look like, and what specific information it might specify?

@blueglyph I honestly think this would be possible, but like you I'm not the most familiar with  stylesheets (or Pelican, for that matter)! Both of the examples you pointed out use [mdbook](https://github.com/rust-lang/mdBook), which have that feature by default. I honestly just searched things up until I got a working search, but at this point I'm not sure if I have the time to make a change as big as that.

I'm thinking that if this were merged, there might be value in creating another issue for moving search to the header because I could imagine users would want to search even if they're not on the main page.

ndd7xv:
This is what it looks like now:

Screenshot 2023-02-11 at 12-42-44 This Week in Rust

nellshamrell:
@ndd7xv Ty SO much for doing this. I'm finally able to turn my attention to this PR.

I am currently getting these errors in the console when running your branch locally:

nellshamrell:
For context - how I'm building the site is 

From the /publishing directory

```
make build && make generate-website && make host-content 
```

I am looking at the local build using Firefox.

nellshamrell:
Interestingly, the search DOES work well in Chrome. It's possible I have a Firefox setting that is throwing it off.

nellshamrell:
Aha! It also works fine in Firefox if I use localhost:8000 rather than 127.0.0.1:8000 - I see the connection here. Trying one more thing...

nellshamrell:
I unfortunately had to revert this merge due to some issues with our email generation process (which you had no way of knowing about, @ndd7xv, no worries at all! I still REALLY appreciate you doing this!)

Details are in #4075, if anyone has ideas on how we might fix it or adapt our flow, let's capture them there :)

andrewpollack:
Hey all! This search functionality is fantastic, and only required a small adjustment to get working with the email workflow. 

Follow-up can be found at #4076

ndd7xv:
Sorry that you had to revert this, I definitely should've tried testing out the newsletter stuff as well! I'll make a note for myself for the future :)  

Thanks @andrewpollack  for the kind words and identifying a fix! In spite of being completely fine if this didn't get merged, I'm grateful that you were able to figure out what was wrong and create a PR :smile:

PR #3943, opened 2023-01-09
Add Fornjot's weekly release

hannobraun:


hannobraun:
Guess we don't need this anymore :smile:

PR #3947, opened 2023-01-10
Add article: Cross compiling Rust with Drone CI

jacobkiers:
This is one out of two posts that describe how to set up Drone CI with Gitea and enable it to cross compile Rust code. This is the post that describes the Rust cross compilation.


U007D (approved):
Thank you for your contribution to TWiR!

PR #3951, opened 2023-01-11
Added OkayWAL announcement

ecton:
I wrote about a new database-related crate I released recently.


U007D (approved):
Thank you for the contribution!  And OkayWAL looks really exciting.  Congratulations, very impressive.

PR #3954, opened 2023-01-11
Add Rust and default parameters

jhartzell42:
This post discusses why default parameters are not included in Rust, and what you can do instead.


U007D (approved):
Thank you for the submission!  Nice article!

PR #3956, opened 2023-01-12
re-add deleted "Rust Walkthroughs" section heading

ericseppanen:
This was accidentally deleted in merge commit 6275fa0fad25d.


U007D (approved):
Thanks for the fix.

PR #3961, opened 2023-01-15
Release of sphrs 0.2.0

stefan-k:
Thanks! :)


U007D (approved):
Thank you for the submission, @stefan-k!  I note that your site does not support `https`; FYI, There may be people who won't/can't visit as a result.

stefan-k:
Thanks for the hint, I'll have a look at it!

stefan-k:
Should now work with `https` as well. Thanks again! :)

PR #3963, opened 2023-01-16
Add why we need type information

h33p:
I may be having a dream, but I can't see if I submitted a PR already or not, even though I recall doing it, but I changed the title to a clearer one (previous one being "We Don't Need Stable ABI")


U007D (approved):
Thank you for the contribution, @h33p!  I love your site's design, too.

PR #3969, opened 2023-01-18
Add event in Kyoto, Japan

asafigan:
I just started a meetup group and this will be its first event. I'm giving the talk.


U007D (approved):
Thank you!  And they look like interesting talks!  Any chance they will be recorded?

asafigan:
The Kyoto talk will not be recorded. I might make an article on the same subject though.

asafigan:
I just realized that I readded the Tokyo event that has already happened when resolving a merge conflict. Oops.

PR #3973, opened 2023-01-19
2D game base with Bevy and LDtk

janos-r:


U007D (approved):
Thank you for the contribution, @janos-r.  Congratulations on all of your hard work!  In the future your posts would be an improved fit for TWiR (and possibly even more helpful to your readers) if you could include some concrete examples showing something you have learned.

PR #3976, opened 2023-01-20
Zenoh 0.7.0 Release -- Rust Pub/Sub/Query Protocol

kydos:
As commented on the rejected pull request, Zenoh is a 100% rust protocol (See https://crates.io/crates/zenoh and http://github.com/eclipse-zenoh/zenoh). We have had the releases covered in the past (see content/2022-10-19-this-week-in-rust.md), not sure why the pull request for the last instalment was rejected, thus re-iterating.

### DIFF
--- a/draft/2023-01-25-this-week-in-rust.md
+++ b/draft/2023-01-25-this-week-in-rust.md
@@ -34,6 +34,8 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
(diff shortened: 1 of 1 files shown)


bennyvasquez:
Hi there! It looks like there was a misunderstanding on your last submission, but we can accept this for sure. One way to help in the future would be to include all of the context that you did here. Sorry for the duplicate and extra work, but thank you very much for making sure we got to share your update!

PR #3983, opened 2023-01-24
add video "10 Reasons Not To Use Rust (The Whole Truth)"

blueglyph:
Ironic title - actually 10 good reasons to use Rust.


bennyvasquez:
We can all use a little levity. :D Thanks for the submission, and congrats on your first TWiR contribution.

blueglyph:
Thanks for all the work you are doing on TWiR!

PR #3984, opened 2023-01-24
Add blog post: "Packaging Rust Applications for the NPM Registry"

orhun:
Hello! 🐻

I wrote a blog post about packaging Rust applications for the NPM registry and automating the process via GitHub Actions. I use the same approach in my project [git-cliff](https://github.com/orhun/git-cliff) which is written in Rust.

I think it would be an interesting addition to TWIR.

Thank you for your consideration!


bennyvasquez:
This one fits better in the miscellaneous category, so I moved it around to there while resolving the conflict. Thanks for the submission!

PR #3985, opened 2023-01-24
Add newsletter: Console #141

orhun:
Hello! 🐻

Recently I had an interview with [Console](https://console.substack.com/p/console-141) (open source newsletter) about my project [git-cliff](https://github.com/orhun/git-cliff) which is a changelog generator written in Rust.
In there, there are a couple of points about Rust and code examples for implementing a minimal changelog generator.

I think it would be interesting for the readers!

Thank you!

### DIFF
--- a/draft/2023-01-25-this-week-in-rust.md
+++ b/draft/2023-01-25-this-week-in-rust.md
@@ -33,6 +33,8 @@ and just ask the editors to select the category.
 
 ### Newsletters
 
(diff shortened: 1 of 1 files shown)


bennyvasquez:
Since we've already got one git-cliff reference in this edition, and this publication isn't rust-centric, we're going to pass on this one. Thanks for the submission, though!

PR #3992, opened 2023-01-25
Add tutorial video on interior mutability

blueglyph:
Interesting tutorial on smart pointers in interior mutability.

It's not complex material so I'm not sure it fits TWiR, but Code to the Moon is a good content creator mainly focused on Rust who we don't see often, so I'll tentatively suggest it.


bennyvasquez:
Thank you for this! And congrats on your first contribution for TWiR!

PR #3997, opened 2023-01-26
Add "Some Rust breaking changes don't require major version" post.

obi1kenobi:
Thanks for all the awesome work you're doing in TWiR! ❤️


nellshamrell (approved):
Ty!

PR #3998, opened 2023-01-26
Added lifetimekata

tfpk:
Should have added this a little while ago, but it's had a couple updates and I figure it's useful to put here now rather than never :)

### DIFF
--- a/draft/2023-02-01-this-week-in-rust.md
+++ b/draft/2023-02-01-this-week-in-rust.md
@@ -39,8 +39,7 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
- - [LifetimeKata: Exercises to learn how to use lifetimes.](https://github.com/tfpk/lifetimekata)
-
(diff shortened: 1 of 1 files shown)


nellshamrell:
This looks really cool!

We generally do not include direct links to github repos (there are so many awesome Rust related projects coming out each week and we cannot include them all), though we do sometimes make exceptions for a README that reads like a blog post.

Would you be willing to either write up a blog post or flesh out the README more? You could include information about why you created this, what inspired it, how it can help others learn Rust, etc.

tfpk:
Hey! I've changed the link to the gitbook rather than the github link. It's more a teaching resource than a "project" in the sense I think you mean; but I see the rationale so hopefully the link change fixes things :)

nellshamrell:
Awesome! Since we already published last week's issue, do you mind moving it to the new draft?

bennyvasquez:
@tfpk it looks like this is still committed to last week's edition. Want to give this another try? If you'd rather, I can just include it in my PR when I am working on this week's issue tomorrow.

tfpk:
Moved to #4029 thanks :)

PR #3999, opened 2023-01-27
Add autometrics 0.1

emschwartz:
Just open-sourced this little library for adding metrics to any Rust function.

Thanks for your work on this newsletter!

### DIFF
--- a/draft/2023-02-01-this-week-in-rust.md
+++ b/draft/2023-02-01-this-week-in-rust.md
@@ -35,6 +35,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
Thanks for the submission.

While we generally do not include direct links to GitHub repos (there are so many awesome Rust projects that come out each week and we cannot include them all), we do sometimes make exceptions when the README is formatted like a blog post with lots of how to's, examples, etc. In the case of this repo, I do believe the README meets that bar, so I am merging this in.

TY again!

emschwartz:
Thank you for including it! I'll make sure to write up a proper blog post for any future releases.

PR #4000, opened 2023-01-28
[Blogpost] One step forward, an easier interoperability between Rust and Haskell

yvan-sraka:
https://engineering.iog.io/2023-01-26-hs-bindgen-introduction/


nellshamrell (approved):
Ty!

PR #4003, opened 2023-01-30
[Walkthrough] Protohackers in Rust, Part 00: Dipping the toe in async…

d2718:
… and Tokio

The first in a series of posts about solving the [Protohackers](https://protohackers.com/) problems with an emphasis on introducing async concepts and Tokio.

Someone like this and suggested I submit it here; thanks for considering it!


nellshamrell (approved):
Ty!

PR #4006, opened 2023-01-31
Add "Moving and re-exporting a Rust type can be major breaking change" post.

obi1kenobi:
Thanks for all the awesome work you do on TWiR!


nellshamrell (approved):
Ty!

PR #4012, opened 2023-02-01
Add Rust bounds checking blog post

jbensal:
Adding in a post on the cost of Rust's runtime bounds checking


nellshamrell:
Looks like we already got this one in another pull request - but ty so much for submitting it too!

PR #4013, opened 2023-02-01
Adding this week's CFPs.

bennyvasquez:
Only one CFP this week. Also a lingering update to the Makefile got brought along that was generally innocuous(just updating the URL for making the email), so I didn't do the dance to un-include it. Hope that doesn't cause problems!


nellshamrell (approved):
Ty!

PR #4015, opened 2023-02-01
docs: add grpc walkthrough to 2023-02-01

shaneutt:
This adds a blog post which walks through using Rust+Tonic to build and test gRPC APIs.


nellshamrell (approved):
Ty!

PR #4020, opened 2023-02-04
add link to EventSourced under project/tooling updates

hseeberger:


bennyvasquez:
Hi there! This looks really cool. 

We generally do not include direct links to github repos (there are so many awesome Rust-related projects coming out each week and we cannot include them all), though we do sometimes make exceptions for a README that reads like a blog post.

Would you be willing to write up a blog post? You could include information about why you created this, what inspired it, how it can help others learn Rust, etc.

hseeberger:
Hi,
That makes sense to me. I already thought about writing a blog post, but I was too lazy. Now I have one more reason ;-)

PR #4021, opened 2023-02-05
Add link to RustyHermit @ FOSDEM 2023 talk

mkroening:


mkroening:
Is this the correct section for such a link?

bennyvasquez:
Yes! You can also add them to a specific section if it applies, but the Miscellaneous section works just fine. Thank you for submitting!

PR #4034, opened 2023-02-08
add FOSDEM 2023 videos in their own section

ericseppanen:
This adds all 17 talk links from the [Rust track](https://fosdem.org/2023/schedule/track/rust/) and another Rust talk from another track that was filed under "Miscellaneous".

I chose to not include any of the talk subtitles for brevity; I'm not sure if that was the right call because some talks have kind of vague titles by themselves.


bennyvasquez:
I think this works quite well, given the space that we're dedicating to it. Thank you for putting it together!

PR #4019, opened 2023-02-03
Add MY talk for next week

pickfire:
Sorry for last minute, haven't plan my time to organize it.

PR #4038, opened 2023-02-08
Add rust malaysia meetup video

pickfire:
We missed the chance to add our event to calendar, so we add it as a post instead.

https://github.com/rust-lang/this-week-in-rust/pull/4019 as mentioned here by @bennyvasquez


bennyvasquez:
Thank you!

PR #4054, opened 2023-02-14
Add event in Kyoto Japan

asafigan:
This is an event that I'm hosting.


andrewpollack:
Looks great, thank you!

PR #4080, opened 2023-02-21
Rust Nederland Embedded Rust event is in Delft not Nijmegen

Kriskras99:


andrewpollack:
Thank you for correcting the location! Merging now

PR #4092, opened 2023-02-23
Some exercises to accompany the Rust Book

tarquin-the-brave:
Hi there!

At hx we've been running a Rust course internally as we adopt Rust as our backend language going forward.  It follows the book and we thought we'd share the exercises we wrote.


mariannegoldin (approved):
Really enjoying these walkthroughs, and the presentation. Thanks for sharing them! Approved.

PR #4093, opened 2023-02-23
Add "rust-analyzer" to the changelog link

teor2345:
It's currently just called "Changelog", which is a bit confusing.


mariannegoldin (approved):
Thank you for this helpful annotation. Approved.

PR #4105, opened 2023-03-01
Article on mailcrab, dev email server in Rust

erikjee:
Short article introducing MailCrab, an email test server for development, written entirely in Rust. 

The backend uses Axum; the frontend is uses Yew.


mariannegoldin (approved):
Thank you for the submission, approving.

PR #4120, opened 2023-03-05
Add blog post/project: "Introducing runst: Handle desktop notifications neatly on Linux!"

orhun:
Hello! 🐻

I just shared [runst](https://github.com/orhun/runst) which is a dead simple notification daemon written in Rust.

I also wrote a blog post about its features and usage examples. This PR simply updates this week's draft about that blog post.

I think it would be interesting for TWIR readers! 🦡

Thank you!


andrewpollack:
An interesting read indeed! Thank you for sharing 😸

PR #4121, opened 2023-03-05
Add blog post/project: "Foogold"

ayrat555:


nellshamrell:
Hello @ayrat555!

Thank you for submitting something to This Week in Rust! 

We've had some internal discussion among the editors on whether on not we should include this link in this week's issue. While we have included material which highlights how Rust can be used to find security vulnerabilities in the past (and recognize the need for offensive security to to reveal holes in systems so that they can be repaired), we don't feel it would be responsible for us to feature this tool in TWiR.

Thank you again for submitting something to us, though!

PR #4132, opened 2023-03-07
Add post about Zellij

imsnif:
Hey - we just released a very significant version of Zellij and would love to be included in the newsletter.


andrewpollack:
Thank you!

PR #4135, opened 2023-03-08
Link to clevercli convert-to-rust

olalonde:


andrewpollack:
Thank you for the submission @olalonde. Unfortunately, this project is lighter on Rust specific info. Our [README](https://github.com/rust-lang/this-week-in-rust#what-do-we-look-for-when-considering-whether-to-include-something-in-this-week-in-rust) gives information on what we are looking for to include:

Updates on tooling when in long form or framed as a tutorial (this can be through a blog, through GitHub, through a newsletter, and more) - it must have a high amount of Rust specific info (examples in Rust, notes on things learned about Rust in the process of creating/updating the project, etc.)

PR #4142, opened 2023-03-09
Typo 2023-03-08-this-week-in-rust.md

jackos:


andrewpollack:
Thank you for the fix!

PR #4175, opened 2023-03-22
fix python urllib syntax

ericseppanen:
On Ubuntu 22.04 with Python 3.10.6, the script crashes with
```txt
AttributeError: module 'urllib' has no attribute 'parse'
```
Apparently the import was wrong all along, but some quirk of the older distro allowed it to work previously.


U007D (approved):
Nice catch, @ericseppanen, thank you!

PR #4188, opened 2023-03-26
Add "Embedded Rust on BBC Micro Bit: unlocking Vec and HashMap"

cyril-marpaud:
Thank you for TWiR =)


andrewpollack:
Thank you @cyril-marpaud for another piece of fantastic embedded Rust content! 😸

PR #4206, opened 2023-03-29
add Paradigm Shift

jhartzell42:
Adding a post I co-wrote with a friend, which contextualizes Rust in a general shift towards functional paradigm approach and programming language features.


andrewpollack:
Thank you!

PR #4210, opened 2023-03-30
Add to virtual events "UA Rust Conference 2023"

ChapliaS:
2023-04-08

### DIFF
--- a/draft/2023-03-29-this-week-in-rust.md
+++ b/draft/2023-03-29-this-week-in-rust.md
@@ -284,6 +284,8 @@ Rusty Events between 2023-03-29 - 2023-04-26 🦀
     * [**Rust Iran Meetup #9 - Let's Write An Async Executor**](https://rust-meetup.ir/2023/04/06/9th-meetup.html)
 * 2023-04-08 | Virtual | [Rust GameDev](https://discord.gg/yNtPTb2)
     * [**Rust GameDev Monthly Meetup**](https://discord.gg/yNtPTb2)
+* 2023-04-08 | Virtual | [Ukrainian Rust Community](https://github.com/rust-lang-ua/learn_rust_together#-ukrainian-rust-community-)
+    * [**UA Rust Conference 2023**](https://www.uarust.com/)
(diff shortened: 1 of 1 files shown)


andrewpollack:
Thank you for this PR @ChapliaS ! I apologize, as this PR will need to be adjusted to target [next week's newsletter](https://github.com/rust-lang/this-week-in-rust/blob/master/draft/2023-04-05-this-week-in-rust.md). This PR came up before I pushed the draft for next week, so timing was on me

Would you be able to re-create this PR, targeting the draft at https://github.com/rust-lang/this-week-in-rust/blob/master/draft/2023-04-05-this-week-in-rust.md ? Apologies, and thank you!

Wandalen:
Hello Andrew! Actually, we were advised to add it to both weeks.

> Not sure why it wasn't mentioned in the newsletter, but you can send them a PR to https://github.com/rust-lang/this-week-in-rust, and they should be able to add it still for yesterday's post (and it will then also be mentioned next week, though that is very close to the conference then).

That is from Jan-Erik Rediger

I personally think that is a better option because there is very little time left. 
Do you mind adding it two both weeks? I will open another PR the next week.

Thanks for the quick response.

Wandalen:
@nellshamrell  30 people invested hours and days of personal time for the common good of the Rust community as well as Ukraine organizing this event. They did it with no commercial incentive. Most people will plan their next Saturday this week, and publishing it three days before the conference beats the whole point of publishing information about it here.

I would not classify it as new content, it's rather a correction, and as you said, you do it occasionally. Maybe you can do it for us too? That's a big difference for us. The difference in the number of participants -> difference in money donated -> saved a life(s). **Even if one life**, it's a huge difference!

nellshamrell:
To discuss further, please reach out to me via email. nellshamrell AT gmail.com

Wandalen:
@nellshamrell sorry you don't care 💔 

@andrewpollack 
Please, add our conference at least to the next week

andrewpollack:
Added: #4213

PR #4232, opened 2023-04-05
Add blog article 'Follow up on cracking ZIP archives in Rust'

agourlay:
Add "Follow up on cracking ZIP archives in Rust" to the "Observations/Thoughts" section.

Thanks for the work on the newsletter!


bennyvasquez:
Thank you!

PR #4243, opened 2023-04-08
Smaller Docker image

apiraino:
Switched to using a base Python image instead of the Ubuntu one (therefore no need to install Python). The new Dockerfile produces a final Docker image of about 2/3 of the size (~387MB instead of ~700mb on my machine).

I tried running all commands in the Makefile and all seems fine (though I am not terribly familiar with the workflow of TwiR).

### DIFF
--- a/publishing/Dockerfile
+++ b/publishing/Dockerfile
@@ -1,15 +1,14 @@
-FROM ubuntu:20.04
+FROM --platform=linux/amd64 python:3.8.16-slim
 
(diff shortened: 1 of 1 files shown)


bennyvasquez:
Thank you, @apiraino! Tagging in @andrewpollack for a review.

andrewpollack:
This looks fantastic @apiraino, thank you! I ran this through our usual TWiR workflows, and all looked solid. With the small merge conflict on our end resolved, this is great to merge 🙌

PR #4265, opened 2023-04-14
Clarify rust-analyzer link title in issue 490

kpreid:
The link should say that it's about rust-analyzer, not just “Changelog” since there are many project changelogs in TWIR. This also matches the style of previous weeks' rust-analyzer links.


andrewpollack:
Thank you!

PR #4269, opened 2023-04-16
Add a Chinese post in the section Rust Walkthroughs

photino:
[Zino's implementation of an error type with tracing functionalities using 100 lines of code](https://zhuanlan.zhihu.com/p/621839043)


nellshamrell (approved):
Approving based on reviewing the post using Google Translate. Normally I would consult our fantastic Chinese language reviewers, but we're a little too close to publish time. I will make sure to do these reviews earlier next week in order to give them enough time to do a review.

PR #4291, opened 2023-04-20
Add Enzyme info

ZuseZ4:
I understand that you usually don't announce MCP, since they are quite a few more than RFCs.
However, I think that this one is a bit bigger than usual and will be followed by an RFC in a year or so.
For the meantime I want to get people interested, in order to have as many test users as possible once I finished the integration into rustc. Could we therefore please announce it here?

I do not really care about the specific design, if you prefer we can remove the MCP section and add it in another place, just let me know!

### DIFF
--- a/draft/cft_rfc_fcp_draft_template.md
+++ b/draft/cft_rfc_fcp_draft_template.md
@@ -5,6 +5,12 @@ are the RFCs that were approved for implementation this week:
 
 * *No RFCs were approved this week.*
 
+#### Approved MCPs
+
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hello @ZuseZ4! My apologies, I forgot to add next week's draft when I published yesterday. Do you mind adding this link to [next week's draft](https://github.com/rust-lang/this-week-in-rust/blob/master/draft/2023-04-26-this-week-in-rust.md) instead?

ZuseZ4:
oh please no worries, thanks for doing the work!
Yes, I can do that. Any thoughts on where to put it?

PR #4344, opened 2023-05-08
Added link to The Way of the Crab ep 1

ecton:
I wanted to ensure that at least the first episode was linked to for the community to discover.

I'm not currently planning on submitting future episodes, since we're currently planning on doing them weekly and that seems like a bit much to include. However, if the newsletter maintainers would actually like me to add each episode we publish, I will happily do so!

Thank you as always for your work behind the scenes!


mariannegoldin (approved):
We are happy to consider PRs for individual new episodes, and thanks for this submission. Approving.

PR #4389, opened 2023-05-25
Fix formatting in 2023-05-24-this-week-in-rust.md

jannic:
Due to an unfortunate line break location, a # character was placed in the first column, causing the line to be spuriously formatted as a headline.


andrewpollack:
Great catch, thank you!

kpreid:
The misformatting is still present on the live <https://this-week-in-rust.org/blog/2023/05/24/this-week-in-rust-496/>.

PR #4410, opened 2023-06-01
Add link to this-week-in-rust.org in the description

briankung:


U007D (approved):
Nice addition, thank you, @briankung.  I'll add this to our process so we retain this addition going forward.

PR #4413, opened 2023-06-04
Video of talk: The status of parallel rustc by Nicholas Nethercote

erikjee:
Video of Nick's talk at our RustNL 2023 conf, about status of parallel rustc.

PS The full conference video was already in TWiR a couple of weeks back, but this talk was not very visible that way, and I'm guessing that there is a lot of interest in this topic. We just published the individual videos on youtube last week.


U007D (approved):
Thanks, @erikjee.

PR #4427, opened 2023-06-08
Add photos.network CfP

thebino:


bennyvasquez:
Hi there! After some further looking here, it seems like this one is not yet appropriate for inclusion. While this has the potential to trigger some interesting discussions, this project is not written in Rust, and this is not a request for something specific to be written in Rust. This section is intended to provide readers with a way to contribute to an open source rust-based project. 

One way to gather the same kind of feedback that I think you're looking for, however, is to write a blog post about why you are considering a rewrite, what you've learned while you have been working on this, and what challenges you are anticipating. Then we can include that in the _Updates from Rust Community_ section, probably under _Observations/Thoughts_.

PR #4431, opened 2023-06-12
Add article about using Kani to check ChatGPT-generated code

CarlKCarlK:
Shows how and when ChatGPT can be trusted. Also, gives an introduction to Kani. Finally, it also gives an intro to Rust's arithmetic functions and the issue of underflow/overflow.


bennyvasquez:
Thank you!

Issue #4336, opened 2023-05-04
Decide where CfP links should go


ericseppanen:
When CfP links come in, should they go under the community section? The "Call for Participation" section? The events section?

Also, do we want CfP links to be sticky? (carried over to the next TWiR issue, until the deadline passes, like events?) If they're sticky, we may need to adjust the CI checks, so they don't cause "duplicate link" errors. And maybe the new-issue/template script too?

If we decide to put them in the "Call for Participation" section, we should probably refine the text header, which only talks about project participation, not conferences or other events:
> Always wanted to contribute to open-source projects but did not know where to start?
> Every week we highlight some tasks from the Rust community for you to pick and get started!
> 
> Some of these tasks may also have mentors available, visit the task page for more information.

mariannegoldin:
I would vote for putting the calls for proposals/speakers under the Call for Participation section, as a subheader - Call for Proposals. I would vote for making the calls sticky for up to four issues: this way, they get enough exposure for readers that may have missed them,  but don't appear on repeat for a long period -- some CfPs deadlines are months out.

This is similar to what we do with events (we publish one month out, meaning that events persist over up to four weeks, allowing readers to catch up on upcoming events). 

I would recommend creating a submission mechanism similar to the Calls for Participation submission [thread](https://users.rust-lang.org/t/twir-call-for-participation/4821/70), via the Rust users forum.

mariannegoldin:
Wanted to see if there were any other thoughts on this thread, vis-a-vis a recent [PR](https://github.com/rust-lang/this-week-in-rust/pull/4341/files) to include a CfP in the general Calls section for TWIR 494.

bennyvasquez:
I agree with keeping them in the Call for Participation section. While topically it would make sense to put them with events, this is a type of participation that is very active. Started to add this in [4435](https://github.com/rust-lang/this-week-in-rust/pull/4435)

PR #4435, opened 2023-06-14
Draft for inclusion of CFPs in the CFPs

bennyvasquez:
Starting to address [4336](https://github.com/rust-lang/this-week-in-rust/issues/4336).

### DIFF
--- a/draft/DRAFT_TEMPLATE
+++ b/draft/DRAFT_TEMPLATE
@@ -51,7 +51,9 @@ and just ask the editors to select the category.
 
 [submit_crate]: https://users.rust-lang.org/t/crate-of-the-week/2704
 
-## Call for Participation
+## Call for Participation; projects and speakers
+
+### CFP - Projects
 
 Always wanted to contribute to open-source projects but did not know where to start?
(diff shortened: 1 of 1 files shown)


bennyvasquez:
Suggested forum post:

> Hello Rustaceans!
> 
> Your [This Week in Rust](https://this-week-in-rust.org/) editors have been discussing a problem that we sometimes encounter where someone wants to submit an event's call for speakers to TWiR, but there isn't a defined place for those. We are happy to share that starting with issue 500 we will be including a new sub-section under the Call for Participation section that is specifically devoted to events that are looking for speakers!
> 
> If you are building an event that would be included in our Events section, feel free to submit your CFP for inclusion below or by submitting a PR on the [This Week in Rust](https://github.com/rust-lang/this-week-in-rust) repo (ensuring you are only including it in the most recent draft).
> 
> Some guidelines to follow while submitting your CFP:
> 
> (not sure what these should be since I couldn't find any defined guidelines for our events section, but I think whatever we write should be the same for both sections)

nellshamrell (approved):
Ty!

PR #4459, opened 2023-06-21
Add blog on why we rewrote Ockam in Rust

mrinalwadhwa:
In [this article](https://www.ockam.io/blog/rewriting_in_rust) I shared Ockam's journey from C to Rust and the things that drove our choice. Hope you would consider including it in this week's newsletter.


bennyvasquez:
This looks great. Thank you!

PR #4499, opened 2023-06-30
Add part 2 of series on understanding async/await

hds:
The focus is on understanding how a pending future gets woken in an async runtime.

Thank you for This Week in Rust. It's a fantastic resource!


nellshamrell (approved):
Ty!

PR #4500, opened 2023-06-30
Add "A persistent task queue in Rust"

jmmv:
This article was published last week but, as it was not included in the previous edition, it could be considered for this one.


nellshamrell (approved):
Ty!

PR #4507, opened 2023-07-04
feat(workshop): add link to devbcn rust workshop

robertohuertasm:
This is a workshop that will take place live July 5h but we (@piny4man and I) prepared an mdbook for everyone to follow. Hopefully it may be useful for other people.


nellshamrell (approved):
Ty!

PR #4528, opened 2023-07-11
Add "Deref specialization in const contexts" to observations

loyd:


mariannegoldin (approved):
Great article, thanks for sharing it. Approved.

PR #4541, opened 2023-07-13
Article on why Rust works for embedded development

erikjee:
A fresh "Why Rust is a great fit for embedded software". We did one before in 2020.

Thanks!


mariannegoldin (approved):
Timely article, thanks! Approving.

PR #4559, opened 2023-07-19
Add Async Rust on Cortex-M Blog Post

bahildebrand:
I recently published this article on running async Rust on cortex-m microcontrollers. I think it might be interesting for TWIR readers.

Thanks!


mariannegoldin (approved):
Thanks for the thorough writeup. Approved.

PR #4579, opened 2023-07-28
Self-nominate "Transition from Go to Rust" to 2023-08-02

HDauven:
Hey! 

I've added a new article to the 'Observations/Thoughts' section on the transition of Dusk's Node software, [Rusk](https://github.com/dusk-network/rusk), from Golang to Rust. I think it's particularly relevant as it provides a concrete example of a project choosing to move a large component of its existing Golang code base over to Rust.


andrewpollack:
Thank you!

PR #4583, opened 2023-07-31
Add video: What to Expect from RustConf 2023

CleanCut:
RustConf is coming right up! It's so exciting! Here's my thoughts on what to expect.


andrewpollack:
It looks like this draft is out of sync with the current -- can you recreate this PR on the current draft?

andrewpollack:
I've bumped this video post to be part of the PR #4593 ! Closing this as we are merging this post as part of that PR

PR #4584, opened 2023-08-01
add part 3 of series on understanding async/await in Rust

hds:
This part looks at what happens when you hold a mutex guard across an await point.

This Week in Rust is great. I never miss an issue!


andrewpollack:
Such a great article, thank you!

PR #4613, opened 2023-08-09
Suggest Enum Optimization with Perfect Hashing

Swatinem:
Explains how the compiler treats big enum matches, and experiments with how to optimize those.

> Optimizing Rust Enum `Debug`-ing with Perfect Hashing

https://swatinem.de/blog/optimizing-enums/


U007D (approved):
Very interesting article, thank you!

PR #4633, opened 2023-08-17
Add actix-contrib-logger

mrsarm:
Hi ! here is a small project I made to be used with the Actix Web framework, it's already published in [Crates.io](https://crates.io/crates/actix-contrib-logger) with documentation of how to use it.

### DIFF
--- a/draft/2023-08-23-this-week-in-rust.md
+++ b/draft/2023-08-23-this-week-in-rust.md
@@ -35,6 +35,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


bennyvasquez:
This looks like a great project! We typically look for blog posts talking about the project, typically focused on why it was necessary and things that you might've learned as you were building it. Do you have an interest in writing something like that and linking there instead?

mrsarm:
Thanks @bennyvasquez !

I tried to explain the motivations in the docs / README of the project, that are mainly a few differences against the original logger middleware.

When having more available time I plan to move some code to new actix-contrib-* crates, that for now is part of another source-closed project, because I think it could be useful for other developers as well, and maybe there I can write a blog post with the motivations for each.

bennyvasquez:
That's a great plan! I'll include this one for now and we look forward to including your blog post in the future.

PR #4657, opened 2023-08-28
Update 2023-08-30-this-week-in-rust.md

c-git:
Content from this week


ericseppanen:
Thanks for the contribution!

It's probably not necessary to submit links to the Rust youtube channel, btw. The editors check there on a regular basis.

We didn't add every video; we reviewed them and picked our favorites.

c-git:
Oh ok great thanks for the info. I'll omit those in future. There were quite a few recently so that will make it easier on me. Thanks for letting me know. I did wonder about it but couldn't figure out where to ask, so just included them in the hope that the duplicate detector would pick it up if someone else also added it. (In hindsight as a comment on the PR might not have been a bad idea to ask if I should in the future)

PR #4662, opened 2023-08-29
Add Zellij release post

imsnif:
Hey - we just released an exciting new version and would love to be included in the newsletter. Thanks!

### DIFF
--- a/draft/2023-08-30-this-week-in-rust.md
+++ b/draft/2023-08-30-this-week-in-rust.md
@@ -34,7 +34,7 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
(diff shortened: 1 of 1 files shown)


bennyvasquez:
Ah, that makes sense, but matching titles isn't at all necessary! If you can cut it down a bit more that would be great! If not (and honestly I'm not sure where you'd cut more at this point), then we'll merge that in and see how it ends up building and looking on mobile.

imsnif:
Alright, I saw this as a challenge. So I dusted off my golfing skills and this is the best I came up with (added a commit). If it's too long still, please feel free to cut out whatever you see fit. :)

bennyvasquez:
That's beautiful! I had to resolve a merge conflict, but now it's good to go! Thank you so much for working with me. :)

PR #4683, opened 2023-09-01
Add article

aochagavia:
Instead of using the original title (Becoming a contractor), I've turned it into Becoming a [Rust] contractor to make clear it may be interesting for the Rust community.


nellshamrell (approved):
Ty!

PR #4687, opened 2023-09-04
Update 2023-09-06-this-week-in-rust.md

Kofituo:


nellshamrell:
Unfortunately I cannot accept this because the full text is available only to Medium members. If you remove the members restriction, I would love to consider including it in a future issue.

PR #4699, opened 2023-09-06
community links for issue 511

ericseppanen:
Also add a trailing slash to an existing link to avoid an unnecessary redirect.


nellshamrell (approved):
Ty!

PR #4712, opened 2023-09-12
Add the blog post "Gotchas in Rust-Python binding with pyo3"

haixuanTao:
This blogpost exposes some easy to make errors when using pyo3 to bridge rust and python. 

It also exposes way to solve those errors.

pyo3 is really good in many regards and I really appreciate the work that have been put in, but there might be challenges people should be aware of. This post does not disrespect the pyo3 maintainers.


mariannegoldin (approved):
This is a fascinating exploration. Thank you, approved.

PR #4724, opened 2023-09-16
add calling Hugging Face models from Rust walkthrough

peterprototypes:


mariannegoldin (approved):
Interesting article, nice to see how the python and Rust envs get combined in a walkthrough. Thanks for submitting. Approved.

PR #4750, opened 2023-09-27
Add Rust Crypto and *ring* blog posts.

briansmith:


nellshamrell:
Hi Brian!

Thanks for these submissions! Because these two posts are very similar, I'm only going to include the second one. It sounds like a very exciting project, though, and I look forward to seeing it develop!

nellshamrell:
And looks like we got that link in with another PR. Closing this, but the post is definitely in the newsletter!

PR #4753, opened 2023-09-28
Add missing event for Twir 514

mariannegoldin:
Add Seattle Rust User Group for 10/19/23, from the events calendar, omitted in error for the 9/27/23 issue.


nellshamrell:
Ty!

PR #4754, opened 2023-09-28
TWIR 514: Link is missing a closing bracket

NTmatter:
The JVM in Rust link is missing a closing bracket, causing the URL to be displayed.


nellshamrell:
Good catch! Ty!

PR #4776, opened 2023-10-05
Add Rust Bern Meetup

dspicher:
Or is this too early? The event is only in November.


andrewpollack:
Thank you for the contribution! Unfortunately, it is a little early to include this on the draft, as we cover events up to one month after publishing date:
```
## Upcoming Events

Rusty Events between 2023-10-11 - 2023-11-08 🦀
```

Please create a PR in a couple weeks once this event falls in the coverage zone!

dspicher:
Right, sorry for the noise!

PR #4796, opened 2023-10-12
Add EtherCrab announcement post

jamwaffles:
This post announces [EtherCrab](https://crates.io/crates/ethercrab). Pretty sure I'm editing the right file!


U007D (approved):
@jamwaffles, you got the right file, indeed!

PR #4798, opened 2023-10-16
Add a link about youki v0.3.0 has been released

utam0k:


utam0k:
@U007D We have categorized and summarized our changelog, not the full changelog. Doesn't it work fine?

PR #4807, opened 2023-10-18
rustc-perf triage for 2023-10-18

pnkfelix:


pnkfelix:
(the last time I posted a twir PR, there were some delays because I used a link that would not be valid until the PR on rustc-perf repo landed. To get around this, this week I used a link that goes straight to the commit id for the draft, which I fully expect to land as is, rather than wait for the PR on rustc-perf to land. Let me know if you all think this is a bad idea.)

U007D (approved):

PR #4808, opened 2023-10-18
Add link to "Writing parsers with Winnow"

adamchalmers:


U007D (approved):
Thank you @adamchalmers!  Love that you're sharing the learning videos that you put together for your employees! ❤️ 
One tip: If you could post in higher resolution, the smaller text for errors, tooltips and the like would be much easier to read.

PR #4812, opened 2023-10-21
Add `Make the Rust compiler 5% faster with this one weird trick` blog post

Kobzol:
https://kobzol.github.io/rust/rustc/2023/10/21/make-rust-compiler-5percent-faster.html (self nomination)


U007D (approved):
@Kobzol approved, thanks!

PR #4814, opened 2023-10-22
add "Associated types in Iterator bounds" in miscellaneous

catalin-i:


U007D (approved):
@catalin-i  approved, thank you!

PR #4815, opened 2023-10-22
add "Nine Rules to Formally Validate Rust Algorithms with Dafny (Part 2)"

CarlKCarlK:


U007D (approved):
@CarlKCarlK approved, thanks!

PR #4817, opened 2023-10-23
Add project update: "ratatui 0.24.0"

orhun:


U007D (commented):
Hi, @orhun.  Per This Week in Rust (TWiR) submission guidelines, we prefer not to link to changelogs (even really, really nice ones like yours) because the content is relevant to a narrow audience.  There are also too many projects which would overwhelm the newsletter with changelogs.

We are happy to post blogs or vlogs relating to `ratatui`.

Congratulations on the new release!

orhun:
Hello, thank you for the feedback. I created #4831 which I think more appropriate 🐻

PR #4823, opened 2023-10-25
add virtual "Rust for Lunch" meetup

hds:
It's the first in a new virtual only meetup at lunch time.


andrewpollack (approved):

andrewpollack:
Thank you!

PR #4850, opened 2023-11-02
Add "Migrating SecureDrop’s PGP backend from GnuPG to Sequoia"

legoktm:
The post is about adopting a Rust library (Sequoia) and explains how/why we're shipping Rust code in a primarily Python project. I'll be honest that I'm not totally sure if this is suitable for TWIR as it focuses a lot on the specific encryption bits rather than Rust-specific things.


bennyvasquez:
Thank you!

PR #4854, opened 2023-11-06
Add post: Using Modern Linux Sockets walk-through

flub:
Hi, I wrote a post about sending from UDP sockets in Rust:
https://devork.be/blog/2023/11/modern-linux-sockets/

thanks


bennyvasquez:
Thank you!

PR #4856, opened 2023-11-07
Add Chitchai Announcement under misc section

ifsheldon:
Hi! I'd like to add an entry under Project/tooling section that introduces my recent project Chitchai. It's a multi-agent chatroom, requiring no paywalls (except subscriptions of GPT providers of users' choice). And it's completely free and open. Chitchai is almost built in Rust, except CSS, using Dioxus.

### DIFF
--- a/draft/2023-11-08-this-week-in-rust.md
+++ b/draft/2023-11-08-this-week-in-rust.md
@@ -36,6 +36,7 @@ and just ask the editors to select the category.
 ### Project/Tooling Updates
 
 * [rust-libp2p v0.53 has been released](https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-v0.53.0)
+* [Announcing Chitchai 0.1: A Free and Private Multi-agent Chatroom in Rust and Wasm](https://chitchai.dev/announcements)
 
 ### Observations/Thoughts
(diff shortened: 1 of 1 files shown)


bennyvasquez:
Hi there, and congrats! Typically we ask for links in this section that are a bit more elaborative or code-focused. Take a look at the posts in the previous Tooling update sections for examples. If you have something like that, we would love to accept it!

ifsheldon:
OK, then I guess it can be moved under Miscellaneous section? 

I didn't mention it in the announcement, but Chitchat can be used as a crate. When I wrote the code, I intentionally tried to make components modular and independent as much as possible so that the UI components can be re-used. In fact, I made the announcement page by reusing chat components. Once the change I made in dependency `async-openai` (to support wasm) gets upstreamed, Chitchat will be ready to be publish on crates.io to be a chat UI component collection. So, I think it's still worth mentioning in TWiR.

bennyvasquez:
Thanks for that additional info! Unfortunately, this still doesn't meet what we'd be looking for. I would be happy to accept an article on how to use it as a crate showing some code examples, or some things that you learned through the development process. 

One other note: We only limit in this way because attempting to cover every single project that is written in Rust would overwhelm the newsletter, but sharing your announcement on Reddit, mastodon, etc would be a great way to expand awareness in the meantime. Keep up the good work!

PR #4858, opened 2023-11-07
Talks on RISC-V + compiler backends-from Rust meetup in Delft NL, Oct 26

erikjee:
Two talks from the https://github.com/rustnl meetup in Delft, NL:

- RISC-V Vector Extension in Rust
- Experiences from building a language and compiler in Rust, and translating those experiences an approach for future improvements to the Rust compiler back-end


bennyvasquez:
Thank you!

PR #4871, opened 2023-11-12
Add my crate `breadcrumbs` to project/tooling updates.

IntegralPilot:
Hi!

I would like to add my new (and first) crate [`breadcrumbs`](https://crates.io/crates/breadcrumbs) to the `Project/Tooling Updates` if that is ok?

Thank you for making This Week in Rust!


nellshamrell (approved):
Ty!

PR #4872, opened 2023-11-13
Add `Cargo Lambda 1.0.0` release to project/tooling updates

calavera:
Hi,

I just released the first major version of [Cargo Lambda](https://www.cargo-lambda.info/), and I'd like to add this highlight to the newsletter.

Let me know if the entry makes sense.

### DIFF
--- a/draft/2023-11-15-this-week-in-rust.md
+++ b/draft/2023-11-15-this-week-in-rust.md
@@ -35,7 +35,7 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hello there! Congrats on the release!

For this section, we are looking for links that are a bit more elaborative or code-focused. Take a look at the posts in the previous Tooling update sections for examples. If you have something like that, we would love to accept it!

We only limit in this way because attempting to cover every single project that is written in Rust would overwhelm the newsletter, but sharing your announcement on Reddit, mastodon, etc would be a great way to expand awareness in the meantime. Keep up the good work!

PR #4888, opened 2023-11-20
Update 2023-11-22-this-week-in-rust.md

szabgab:


nellshamrell:
Thank you for submitting this! I'd like see this article be longer and have a little more substance in order to feature it. How would you feel about combining the old article and this one into one?

PR #4890, opened 2023-11-20
Update 2023-11-22-this-week-in-rust.md

szabgab:


nellshamrell:
Hello there, thanks for submitting this! Typically we ask for links in this section that are a bit more elaborative or code-focused. Take a look at the posts in the previous Tooling update sections for examples. If you have something like that, we would love to accept it!

PR #4918, opened 2023-11-28
Add a post about setting up AWS Lambda functions in Rust with Terraform

ador:


mariannegoldin (approved):
Great article -- appreciate that it covers the deployment story. Approving.

PR #4920, opened 2023-11-28
Add: "SALT: A Rust error tutor"

mojeanmac:
Adds SALT (Situationally Adaptive Language Tutor) to tooling section. Current and future tools as part of SALT can still be accessed without opting-in to participate in the study.


mariannegoldin (approved):
Interesting project, thanks for submitting. Approved.

PR #4925, opened 2023-11-29
Semantic fuzzing of the Rust compiler and interpreter

BD103:
[This](https://ethz.ch/content/dam/ethz/special-interest/infk/inst-pls/plf-dam/documents/StudentProjects/MasterTheses/2023-Andy-Thesis.pdf) is a research paper that I saw on Hackernews a few days ago. It was published in July so I don't know if it's been on This Week in Rust previously, but I figured I would try adding it anyways.

Thanks :)


mariannegoldin (approved):
Thanks for submitting this. Approved.

PR #4930, opened 2023-11-30
feat(meetup): 12th bcnrust meetup

robertohuertasm:


mariannegoldin (commented):
Thanks for the submission, however this issue of TWIR only includes events thru Jan 3 2024. We will note this event internally, but just in case, please re-submit this in a month. Thank you!

PR #4935, opened 2023-12-05
Add an article how to set up a Zettelkasten with Tp-Note (written in Rust)

getreu:
This blog post shows how to set up Niklas Luhmann's Zettelkasten with Tp-Note and how to work with it. Tp-Note is a free note taking tool written in Rust.

### DIFF
--- a/draft/2023-12-06-this-week-in-rust.md
+++ b/draft/2023-12-06-this-week-in-rust.md
@@ -34,6 +34,7 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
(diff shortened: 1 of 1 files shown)


mariannegoldin (commented):
This is a great project, however there is no mention of Rust in the writeup, which would make the topic more relevant to our readership. We would love to see a version of this article with more specific Rust content and would encourage a revision and re-submission for a subsequent issue.

getreu:
Thank you for your feedback. Do you think adding a chapter about the reasons why I chose Rust for this project (and why it was a good choice) would interest your readers?

PR #4943, opened 2023-12-06
add: Diverging Functions - functions that never return

szabgab:


mariannegoldin (approved):
Useful post, thanks for submitting. Approved.

PR #4949, opened 2023-12-08
updated with 524's data

bennyvasquez:


bennyvasquez:
ah, just realized the email content was omitted because of the git config. here you go! 
[out.txt](https://github.com/rust-lang/this-week-in-rust/files/13607916/out.txt)

mariannegoldin (approved):
Ty!

PR #4950, opened 2023-12-08
Rust Is Beyond Object-Oriented, Part 3: Inheritance

jhartzell42:
Rust Is Beyond Object-Oriented, Part 3: Inheritance


andrewpollack:
Thank you!

PR #4952, opened 2023-12-10
Add Watchexec Library 3.0 and CLI 1.24

passcod:
<https://cohost.org/watchexec/post/3818997-watchexec-library-3>

Project update but not just release notes, it's instead a lot of background, specific rust code examples callouts, and thanking a contributor :)


andrewpollack:
Thank you!

PR #4953, opened 2023-12-10
Add "Contributing to Rust as a novice"

aochagavia:
This was originally published a couple of months ago, but I forgot to submit it to TWIR, so here it is at last!


andrewpollack:
Thank you!

PR #4984, opened 2023-12-20
feat: Add my embedded-hal driver article to TWIR 526

onkoe:
This article revolves around my past two years of learning Rust,
so it's pretty important to me!

Please let me know if there's anything that needs to be changed. 😄


andrewpollack:
Thank you!

PR #4991, opened 2023-12-21
Make survey link clearer

teor2345:
It's hard to tell that "Launching State of Rust Survey" and "Survey" are two different links.


andrewpollack:
Thank you!

PR #4997, opened 2023-12-24
Add `static mut` survey to 527

dyslexicsteak:
I created a survey to go with my proposal on IRLO [Pre-RFC: Deprecate then remove `static mut`](https://internals.rust-lang.org/t/pre-rfc-deprecate-then-remove-static-mut/20072), and I'd like to add it to TWiR to get it some more exposure. The survey is entirely anonymous as guaranteed by SurveyHero: "we will hide email address, IP address, link parameters, meta data and any other piece of information that might allow you to identify a participant".


U007D (approved):
Thanks, approved!

PR #5000, opened 2023-12-25
add Web development in Rust

szabgab:


U007D (approved):
Nice project, documenting so many different Rust web frameworks!  Thanks, approved.

PR #5011, opened 2023-12-28
Publishing 527

bennyvasquez:
published!


bennyvasquez:
@U007D here's the output for the email, assuming everything stays the same for tomorrow. 

[out.txt](https://github.com/rust-lang/this-week-in-rust/files/13783122/out.txt)

PR #5012, opened 2023-12-28
Fix a tiny formatting error in This Week in Rust 527

jihchi:
### [Preview content](https://github.com/jihchi/this-week-in-rust/blob/patch-3/content/2023-12-27-this-week-in-rust.md)

One character change.


U007D:
Thank you, @jihchi!  I've made the change myself, but definitely appreciate the PR! 🙏🏾

PR #5017, opened 2024-01-01
add Test command line application written in Rust

szabgab:


U007D (approved):
Looks good, @szabgab, thank you.

PR #5031, opened 2024-01-04
Add tutorial on building a desktop UI app in rust

hunger:
I really liked that video walking you through writing an UI application to having it on the Mac desktop.

Disclaimer: I work on the crate being used in that video, but it was done without any of us aware of its production.


bennyvasquez:
Thank you!

PR #5037, opened 2024-01-06
Add link to "Elevate your CLI experience: Nushell" article

practicalrs:


bennyvasquez:
Hi there, and thank you for your submission! In the "this week at rust" crew our goal is to focus on content that involves coding in rust to some degree. Unfortunately, this seems to focus on using the terminal, rather than rust code specifically. If you follow a similar pattern but focus on something more rust-specific this would be a great article to include. Thank you!

practicalrs:
The title of my next article is "Rust-Infused Compilers: Nushell". I'm trying to deep dive into Nushell interpreter implementation. I hope it will be more suited.

PR #5047, opened 2024-01-10
Add blog post: "Why stdout is faster than stderr?"

orhun:
I'm not sure if the "research" category is correct. It might also be observations/thoughts.

Just let me know!


bennyvasquez:
Thank you! I do think this will fit better in observations, so I'll move it there when I do the final cleanup ahead of publishing.

PR #5061, opened 2024-01-11
add `Adventures in Binary Serialisation` to next week's observations

BurntNail:
Not 100% sure if this is more of a walkthrough or more of an observation - I've put it in the latter, but I'll happily move it to the former.


bennyvasquez:
Thank you!

PR #5121, opened 2024-01-31
adding a "what if there are no CFPs" text to the draft template

bennyvasquez:
Small change that'll give folks something easy to use in the event that we have no calls for papers/participation for events.


nellshamrell (approved):
Excellent! Ty!

PR #5156, opened 2024-02-08
Add January 2024 Rust Jobs Report

drewcapener:
Just adding a link to the January 2024 Rust Jobs Report. I'm not sure if Miscellaneous is the right section, so please move it as needed!


mariannegoldin (approved):
This is great and the correct section, ty! Approved

PR #5161, opened 2024-02-10
Add "Which red is your function?"

JarredAllen:
Some thoughts I have about a minor pain point with async Rust that I hit from time to time (inspired by the recent without boats post, as I'm sure could have been guessed).

Sorry for the lack of a proper blog like most of these posts are, I haven't gotten around to setting a blog up yet, but I had an epiphany about how to explain myself and wanted to get this written down while the ideas are still fresh in my head.

https://gist.github.com/JarredAllen/6cd2fd5faead573d1120a96135ed3346


mariannegoldin (approved):
Interesting and original analogy! TY for this submission, approved.

PR #5163, opened 2024-02-12
Remove AA's case study from TWIR#533

LukeMathWalker:
Hey!
AA asked us to review some of the wording in the case study, therefore we've (temporarily) taken it down and the link now redirects to Mainmatter's Rust homepage.
Since it'd be inappropriate to have a link to Mainmatter's homepage on Twir, I've opened this PR to take it out entirely. If there's a better course of action, let me know!

Sorry again for the inconvenience.


mariannegoldin (approved):
Got it! Thanks for making the update, this makes sense. Approved.

PR #5164, opened 2024-02-12
add Devoxx PL 2024

bbucko:
Added CFP for [Devoxx PL ](https://devoxx.pl/). I'm a member of a program comitee and although Devoxx PL is not a Rust conference - we are focused on Java and programming in general - but I'm interested in pulling more Rust content.


mariannegoldin (approved):
TY for the submission, we appreciate it. Approved.

PR #5169, opened 2024-02-13
Add article

aochagavia:
The SAT solver mentioned in the article is written in Rust, which IMO makes the article relevant for the community.


mariannegoldin (approved):
Thank you, loved reading this. A part 2 with the new directions you proposed would be a welcome submission as well, should you decide to go that route. Approved.

PR #5170, opened 2024-02-13
Blog post on benefits of Rust for hardware vendors

erikjee:
Blog post pitching Rust support to hardware vendors, by [@diondokter](https://github.com/diondokter)

No code in the post, but a lot of arguments for Rust :)


mariannegoldin (approved):
Thanks for the submission and the annotation about the article, although we usually prefer to see code examples, this feels targeted and relevant. Approved.

PR #5171, opened 2024-02-13
Add blog post on Scaphandre v1.0 release

bpetit:


mariannegoldin (commented):
Thank you for this submission, however we are not able to approve at present as this blog post does not contain Rust-specific content, which we prioritize for our readership. We would encourage a resubmission with added Rust context, examples, approaches, architecture, etc. Closing the PR.

PR #5172, opened 2024-02-13
Add Rust FFI in Go blog post

davidmytton:
Adds https://blog.arcjet.com/calling-rust-ffi-libraries-from-go/


mariannegoldin (approved):
Fascinating, and thanks for including resources around c libs and zig. Approved.

PR #5175, opened 2024-02-13
add "Why Rust? It's the safe choice" article link

ericseppanen:
This is me wearing an "author" hat rather than my "editor" hat.  I'm submitting an article I wrote for the Matic Robots company blog.


mariannegoldin (approved):
Thanks for the submission - interesting look into robotics and Rust. Approving.

PR #5195, opened 2024-02-17
Move Call for Testing section higher to increase visibility

Urgau:
This PR moves the "Call for Testing" section higher to be just after the "Crate of the week".

This is done in response to the Call for Testing done this week (534) about check-cfg that received almost no responses; asking some peoples if they had seen it revealed that none of the people I asked had seen it.

I attribute this to the section being buried deep between big blobs of text.

| Before | After |
|--------|-------|
| Screenshot 2024-02-17 at 18-33-07 This Week in Rust 534 · This Week in Rust       | Screenshot 2024-02-17 at 18-47-21 This Week in Rust 534 · This Week in Rust - MOVED      |

### DIFF
--- a/draft/DRAFT_TEMPLATE
+++ b/draft/DRAFT_TEMPLATE
@@ -51,6 +51,25 @@ and just ask the editors to select the category.
 
 [submit_crate]: https://users.rust-lang.org/t/crate-of-the-week/2704
 
+## Call for Testing
+
+An important step for RFC implementation is for people to experiment with the
+implementation and give feedback, especially before stabilization.  The following
+RFCs would benefit from user testing before moving forward:
+
+<!-- Calls for Testing go here, use this format:
+    * [<RFC Topic>](<RFC URL>)
+        * [Tracking Issue](<Tracking Issue URL>)
+        * [Testing steps](<Testing Steps URL>)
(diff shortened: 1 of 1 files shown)


Urgau:
cc @nellshamrell @llogiq any objections or suggestion regarding this change ?

llogiq:
From my point of view, this makes sense, and there's nothing standing against it.

U007D:
This makes sense to me as well.  I'll make the change for tonight's issue and republish the Call for Testing from last week.  Hopefully we see more traction/uptake.

cdmistman:
closing in favor of #5202. I'll update the template as well to ensure that we keep this positioning in the future. thank you @Urgau!

U007D:
Hi, @Urgau,

Out of curiosity, did you see any more activity from the updated posting last week?

Urgau:
@U007D Unfortunately no. We had 2 or 3 replies[^replies] from the previous Call for Testing, but so far we had no further replies.
I'm not sure how to interpret it, maybe peoples are shy and had nothing more to add? (just speculation)

Anyway, thanks you for help.

[^replies]: that I'm relatively certain were coming from the Call for testing and not from other means

PR #5206, opened 2024-02-21
add good first issues from hyperswitch

Narayanbhat166:
Adding few of the good first issues from [hyperswitch](https://github.com/juspay/hyperswitch) repository


cdmistman (approved):
thank you!

PR #5213, opened 2024-02-26
suggest runi

thor314:


cdmistman:
Thank you so much for contributing! While we greatly appreciate unique works like these, we prefer not to simply link to crates.io listings. That said, if you have a write-up about your work in some format (white paper, blog post, walkthrough of using your crate, etc) then we'd gladly share that work!

PR #5225, opened 2024-02-27
Added 'How Rust Could Change Robotics'

drewcapener:
Just adding an interview about Rust in robotics. I'm pretty sure it's in the right spot, but feel free to move it!


cdmistman (approved):
very interesting read! thank you!

PR #5228, opened 2024-02-28
add Rust, but async post

hds:
The blog post looks at what a variant of Rust that is only async might
look like.

I love This Week in Rust, thank you for all your work!


cdmistman (approved):
very interesting thoughts (especially as a fan of PL design/theory), thank you!

PR #5262, opened 2024-03-11
Adds "From medical doctor to rust developer"

briankung:
From medical doctor to rust developer - interview with our new Senior Clinical Research Fellow (Caroline Morton)


nellshamrell (approved):
Ty!

PR #5281, opened 2024-03-14
community links for issue 538

ericseppanen:
Some links moved around during deduplication.

I also replaced some slow url-shortener links with the canonical URLs.


bennyvasquez:
Thank you!

PR #5289, opened 2024-03-16
Add `What part of Rust compilation is the bottleneck` blog post

Kobzol:
Self-nomination for my [blog post](https://kobzol.github.io/rust/rustc/2024/03/15/rustc-what-takes-so-long.html).


nellshamrell (approved):
Ty!

PR #5363, opened 2024-04-05
Add link to Embedding the Servo Web Engine in Qt

ahayzen-kdab:
This is a blog about using CXX-Qt (Rust bindings for Qt) to embed Servo (a Rust web rendering engine) in a Qt application


mariannegoldin (approved):
This is interesting! For future submissions, I would have loved to know a little more about  the details of the Rust implementation (e.g. `surfman` in this case). Approving.

PR #5367, opened 2024-04-07
add event: "Using AstroNvim for Rust development (in Hebrew)"

szabgab:


mariannegoldin (approved):
Thank you for the submission. Approved!

PR #5368, opened 2024-04-07
add "Rust Meetup and user groups (updated)"

szabgab:
This was included once when I originally published it. I've updated the list and the numbers. It might be valuable to share it again.


mariannegoldin (approved):
Thank you, this list is really helpful and we appreciate the update. Approved.

PR #5385, opened 2024-04-11
TWIR Events #542: Change Charlottesville event state to VA, US

mariannegoldin:


mariannegoldin:
TY @JoelMarcey for catching this

PR #5392, opened 2024-04-14
add event "Rust at Microsoft, Tel Aviv - Are we embedded yet?"

szabgab:


mariannegoldin (approved):
Thank you for the submission! Approved.

PR #5400, opened 2024-04-16
Add Zellij v0.40 release post

imsnif:
Hi this-week-in-rust,

We'd love to be included in the next newsleter.

Thanks!


cdmistman (commented):
Hey! While we appreciate your work (I'm personally a fan of zellij) and your submission, we prefer to have project/tooling updates be for things like libraries or very targeted rust tooling like rust-analyzer. As an example of how this might be applied to Zellij (without knowing much about the internals), if you were to share the layout logic via a crate and share an announcement post of some sort with an example usage of the crate, we would love to have that!

PR #5402, opened 2024-04-16
Add Announcing async-openai-wasm

ifsheldon:
Hi this-week-in-rust,

I recently released a crate `async-openai-wasm` and wrote a blog that announces it and how I make it work on WASM. I'd like it to be included in this week's release. The blog has both CN and EN versions.

### DIFF
--- a/draft/2024-04-17-this-week-in-rust.md
+++ b/draft/2024-04-17-this-week-in-rust.md
@@ -34,6 +34,7 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
+* `[ZH][EN]` [Announcing async-openai-wasm, and thoughts on wasmization and streams](https://ideas.reify.ing/en/blog/announcing-async-openai-wasm/): Async Rust library for interacting with OpenAI's APIs on WASM and how I did it.
 
 ### Observations/Thoughts
 


cdmistman (approved):
very neat, thank you!

holly-hacker:
Hi, I recommend reversing this merge. This crate will inevitably cost people a lot of money as it does not have any method of keeping your OpenAI API key secret (and it can never do so, as it runs exclusively clientside). Any bad actor with basic DevTools knowledge can take just take the key and use it to rack up massive bills.

I also suggest adding a large disclaimer to the github repo and the crate description, so people are aware of this.

ifsheldon:
@holly-hacker I think this is beyond what this crate should cover. How key secrets are stored is not the concern of this crate. Just like the official `openai` does not bundle a vault for keys to prevent hackers. There are many ways to secure secrets on the clientside. The most trivial one is to proxy requests with one-off api keys and proxy servers, which is orthogonal to any wasm technologies.

holly-hacker:
Regardless of how you think you can secure API keys clientside, it'd be irresponsible to not at least leave a disclaimer to inform people that API key cannot be securely stored clientside (be it in code, localstorage, or otherwise). Doing these API calls client-side is generally a terrible idea security-wise and should only be done if you have a very good reason, sufficient understanding of the risks involved and a solid plan on how to mitigate them.

PR #5416, opened 2024-04-18
Add zed blog post: language extensions via wasmtime + tree-sitter

maxbrunsfeld:


bennyvasquez:
Hi there! Thanks so much for taking the time to submit this article. Based on my reading of it, it doesn't seem to be rust-centric in content, so it's not quite right for our newsletter. If that's incorrect or you have any other questions, please do feel free to reach out!

PR #5424, opened 2024-04-23
Add "Writing ergonomic async assertions in Rust" blog post

niklaslong:
I've included it under "Observations/Thoughts" for now, hopefully that's appropriate. 

Alternatively, if you'd prefer it to be under tooling updates, the crate name could feature more prominently in the link title, something like: "Deadline: writing ergonomic async assertions in Rust".

### DIFF
--- a/draft/DRAFT_TEMPLATE
+++ b/draft/DRAFT_TEMPLATE
@@ -37,6 +37,8 @@ and just ask the editors to select the category.
 
 ### Observations/Thoughts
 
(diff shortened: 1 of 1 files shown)


bennyvasquez:
That works just fine. Thank you!

niklaslong:
@bennyvasquez just realised I made the change on the wrong file 🤦 we should revert this and I'll adjust the PR so it can be added no. 545. Sorry about that.

PR #5427, opened 2024-04-23
add "Rust Digger: Does size matter? The biggest crate is 450MB; More …

szabgab:
…than 100 crates are over 50MB."


bennyvasquez:
Thank you!

PR #5432, opened 2024-04-24
add article "drop, the destructor of Rust structs"

szabgab:


cdmistman (approved):
nice, i'll move this to the walkthrough section :) thank you!

PR #5439, opened 2024-04-25
Add skipdb release to project/tooling updates

al8n:


cdmistman (changes requested):
this is a very neat project - thank you for sharing! unfortunately we prefer to avoid linking directly to READMEs/repos. If you write something a little closer to the blog post format, where you might explain how your project works, what you learned writing it, and/or examples of how to use it effectively in rust code, we will gladly share that content!

PR #5463, opened 2024-05-01
add post on rust performance optimization

hds:
Using the (cargo) flamegraph tool for profiling and Divan for
benchmarking.

We love This Week in Rust!


bennyvasquez:
Thank you!

PR #5483, opened 2024-05-08
reduce double-link to single link

ericseppanen:
We try to follow a one-link per line rule in the community sections. Remove the second link.


bennyvasquez:
Thanks for this, too!

Issue #3592, opened 2022-08-26
Header formatting on email version of TWiR is not line breaking correctly


mariannegoldin:
The title image and header of the email are not breaking correctly in the generated HTML code for the TwiR email issue. 

The break should be placed after the "This Week in Rust" title image. Instead, the line is currently breaking after the "This" in the text header.

The email code is being generated by the `optimize-email` Makefile job, which uses the website code (via the `make build && make generate-website && make host-content` command)  to create an email using `juice`. 

```
                                    <td style="font-family: 'Helvetica Neue', Helvetica, sans-serif;">
                                      <img class="wrapper-image " src="https://this-week-in-rust.org/images/logo32.png" alt="This Week in Rust" style="width: 30px; vertical-align: bottom; max-width: 100%;" width="30">
                                      <span class="wrapper-header-01" style="line-height: 27px; font-size: 23px; font-weight: bold; color: #444444;">This Week in Rust</span>
                                    </td>
```

Narrow window:
![Screen Shot 2022-08-25 at 6 25 39 PM](https://user-images.githubusercontent.com/23177337/186796760-0dee90c4-8c91-4a31-9a14-74cf1da3d6b3.png)

Wide window:
![Screen Shot 2022-08-25 at 6 23 48 PM](https://user-images.githubusercontent.com/23177337/186796587-138edb73-904e-491d-bf31-2da686bf195d.png)

ericseppanen:
It looks good for me, but I am seeing an icon at the left where you get "This Week in Rust" in small purple text.

Gmail/Linux/Firefox:
![twir_gmail_firefox](https://user-images.githubusercontent.com/36317762/186829927-f69498dd-d0bf-440f-9ed6-9b8fd230d0f6.png)

Gmail/IOS:
![twir_gmail_mobile](https://user-images.githubusercontent.com/36317762/186829974-9a4a1123-4afb-4f48-99bf-5ddf4f78b9e4.png)

ericseppanen:
I think the alt-text on the icon is actually hurting here (since we repeat that exact text to its right), and suppressing it would give a better outcome.

bennyvasquez:
@mariannegoldin / @ericseppanen do either of y'all have interest in trying to fix this? We don't wanna suppress the alt-text, but that's definitely a less than optimal experience for users. If you don't have time or expertise, I'll add that `help wanted` label and we'll see if we can get some help.

ericseppanen:
[...]

PR #5490, opened 2024-05-09
Fixing 3592

bennyvasquez:
Theoretically, this should fix the image display issues in #3592. I've confirmed that the email for 546 generated correctly.


mariannegoldin (approved):
Looks good, ty! Approved

PR #5498, opened 2024-05-12
add event  "Web development in Rust using Rocket (online, Hebrew)"

szabgab:


mariannegoldin (approved):
Thank you! Approving.

PR #5527, opened 2024-05-20
Add Devoxx UK Rust talk to 05-22

Mattsi-Jansky:
As per title, a recording of my talk at Devoxx UK about Rust. Delves deep into explaining ownership and memory models, intended as an introduction into what is interesting and different about Rust.


nellshamrell (approved):
Ty!

PR #5583, opened 2024-06-04
add Rust Frankfurt/DE Meetup

christophbeberweil:
Hi all, Rust Frankfurt hosts a meetup after a long break. It would be great if you could include it in the newsletter :)

Thanks 👋


mariannegoldin (approved):
Thanks for submitting! Approved with a correction for the group link.

PR #5584, opened 2024-06-05
Adds "On Dependency Usage in Rust"

briankung:


cdmistman (approved):
very interesting read, thank you!

PR #5599, opened 2024-06-10
Add "Master Hexagonal Architecture in Rust" walkthrough

AngusGMorrison:
This walkthrough kicks off a series breaking down the implementation of hexagonal architecture in Rust by examining an application with serious maintainability problems.

By exploring what _doesn't_ work, Rustaceans will learn about the problems that hexagonal architecture exists to solve.


mariannegoldin (approved):
TY! Approved

PR #5603, opened 2024-06-11
add "Virtual Rust Events"

szabgab:


mariannegoldin (approved):
Thanks! I appreciate the proactive addition of this. We can iterate over time to find a way to address the local time aspect of the events list that comes from TWIR. Approved.

PR #5619, opened 2024-06-14
Add subscription message to newsletter body

mariannegoldin:
I realized that we didn't have the subscription message in the body of our email. This PR adds it to the draft template + the current #552 of TWIR.


bennyvasquez:
Looks good to me, with one caveat: In the open source communities I'm part of URL shorteners are super frowned on. How does the Rust community feel about them in general?

mariannegoldin:
Good point! We can point this to the fully expanded destination link.

I’ll update the PR.

bennyvasquez (approved):
Looks good to me!

PR #5624, opened 2024-06-17
add "Rust Meetup and user groups" with updated list and numbers

szabgab:
This was already included 3 months ago. I've just updated the list and the numbers in case you'd like to include it.


szabgab:
I'll have another post instead of this.

PR #5625, opened 2024-06-17
July meetup for Reading UK group

timabell:
Update the Reading UK meetup links

- I've got a new page on the site for the meetup which is nicer than the the meetup page so have updated the group link to point to that.
- I've updated the meetup link to this month's event on meetup.com
- I've added "July" to the name to make it a bit clearer.

Let me know if that's okay. A couple of people told me they found the group through twir, many thanks for that! And thanks for a great newsletter.

This is my first PR to twir hope I've done it right!

### DIFF
--- a/draft/2024-06-19-this-week-in-rust.md
+++ b/draft/2024-06-19-this-week-in-rust.md
@@ -197,8 +197,6 @@ Rusty Events between 2024-06-19 - 2024-07-17 🦀
     
 ### Europe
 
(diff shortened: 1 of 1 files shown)


mariannegoldin (approved):
Thank you for your submission! This looks great, and kudos on the infra for your meetup; totally ok to link to the group page. Approved with a re-ordering (events are ordered chrono, ascending, per section). Please keep the submissions coming, as they ensure that the events don't fall through the cracks on our end.

PR #5632, opened 2024-06-19
add event "Web development in Rust using Rocket - part 2"

szabgab:


mariannegoldin (approved):
Thank you for the submission, approving

PR #5654, opened 2024-06-25
add post on Tokio waker tokio-waker-instrumentation

hds:
This is a somewhat in depth look into what we can determine about task
depedencies related to wakers by looking at the output of the tracing
instrumentation in Tokio.

I do love This Week in Rust, thank you for working on it!


bennyvasquez:
Thank you!

PR #5662, opened 2024-06-26
2024-06-26: Add Announcing Summer of Rust 2024

sylvain101010:


bennyvasquez:
This just indicates that you're starting the newsletter, but doesn't provide any rust-related content. Happy to have you share future content that is more than the announcement. Thank you!

PR #5672, opened 2024-06-28
tooling update: what-the-time v1.0.0

sdball:
what-the-time `wtt` is a new Rust tool I've made that replaces a few Ruby scripts with one command.


bennyvasquez:
Thank you!

PR #5673, opened 2024-06-29
Add "The Minimal Rust-Wasm Setup"

dzfrias:
A walkthrough on how to setup up Rust and Wasm with as little external tooling as possible, but still featuring `wasm-bindgen`.


bennyvasquez:
Thank you!

PR #5675, opened 2024-06-30
Intro to Danube - a pub/sub message broker

danrusei:
a new pub/sub message broker written in Rust, inspired by Apache Pulsar.


bennyvasquez:
Thank you!

PR #5677, opened 2024-07-01
Series of articles on interop with rust

erikjee:
Series of articles on interop with rust, starting with two articles called "Mix in Rust" and "Mix in Rust with C".
Thanks!

### DIFF
--- a/draft/2024-07-03-this-week-in-rust.md
+++ b/draft/2024-07-03-this-week-in-rust.md
@@ -41,6 +41,8 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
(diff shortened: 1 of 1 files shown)


bennyvasquez:
Thanks for your submission! Unfortunately, I can't accept this submission as is. We don't link to series like this, only to individual articles. I also wanted to mention (in case you re-submit), we only accept a single link from an author each week.

erikjee:
> Thanks for your submission! Unfortunately, I can't accept this submission as is. We don't link to series like this, only to individual articles. I also wanted to mention (in case you re-submit), we only accept a single link from an author each week.

Thanks for explaining, I didn't know that. I'll re-submit the first article for next week.

PR #5689, opened 2024-07-03
2024-07-03: add It's time to abolish the builder pattern in Rust

sylvain101010:


bennyvasquez:

PR #5691, opened 2024-07-03
2024-07-03: Add Rust Patterns: Make a reference &'static with Box::leak

sylvain101010:


bennyvasquez:

PR #5699, opened 2024-07-03
Fix two incorrect FCP dispositions in TWiR 553 (2024-06-26)

kpreid:


bennyvasquez:
Hi there! Thanks so much for this. The statuses were accurate for the release of the newsletter, so I'm going to decline this update. Thanks for looking out!

PR #5710, opened 2024-07-08
Update 2024-07-10-this-week-in-rust.md

mssabr01:
Added link to Shardeum's bug bounty program. It has some significant pieces built in rust and in need of review


nellshamrell:
Hello there! Thank you for submitting this. Unfortunately, the events section is intended for things like Rust meetups and conferences, not for launches like this. Additionally, there is nothing Rust specific in the announcement, so I unfortunately need to close this.

PR #5724, opened 2024-07-10
Add  thoughts entry for "A Unified Typesystem"

Aras14HD:
I asked for this to be included before on Mastodon, where I was told, you don't accept repos, which this isn't, it's a gist. After I responded with that I got neither a positive nor negative reply. So I will try my luck here.

I think this fits in the this week rather small Observations/Thoughts section.


nellshamrell (approved):
Ty!

PR #5732, opened 2024-07-14
Add blog post on quickwit scaling to 100PB

fmassot:


nellshamrell:
This is very cool. Unfortunately, even though this highlights the use of a product written in Rust, it does not feel Rust specific enough to include in the newsletter. I would love to see a blog post with Rust code samples/crate examples/etc. showing the use of Quickwit (or how Quickwit is constructed). We would love to include something like that in a future issue!

fmassot:
@nellshamrell makes sense, thanks for the feedback!

PR #5736, opened 2024-07-16
Add annoucement about cratery being open-sourced

woutersl:


nellshamrell (approved):
Ty! Merging, though I may shorten the description prior to publish.

PR #5739, opened 2024-07-16
Add link to 'Surprises with Rust's as' to Walkthroughs

anna-hope:
Added my recent blog post about potentially surprising behavior when using `as` to cast types, and why I try to avoid it and prefer `TryFrom` instead. Also includes background information, notably an exploration of Python's division operations, to show why Rust's strict type system might prevent surprising behavior or bugs.

I *think* this fits into Walkthroughs, but might also fit in "Observations/Thoughts" instead. Please let me know if I should move it!


nellshamrell (approved):
This is great! Ty!

PR #5742, opened 2024-07-17
2024-07-17: Add How to organize large Rust codebases

sylvain101010:


nellshamrell (approved):
Ty!

ijackson:
Hi @sylvain101010 .  I'm having terrible trouble reading what I presume to be your blog posts.

My Firefox is getting "403 Forbidden" from your webserver, for some reason.  I tried bypassing my squid proxy but to no avail.

w3m gets me a messadge complaining that JS is needed.  I tried looking at the source code with curl and that got me "Forb idden" again.  I used w3m to dump the "source" and it's some kind of Javascript blobule?

I may be some kind of curmudgeon, but as a security expert, I think asking people to download and run programs on their computer, in order to read a blog post, is not great.  (Ideally the TWIR editors would check that links are accessible, but I quite understand that they probably don't have the human bandwidth to check for that.)

PR #5749, opened 2024-07-19
Fix approved RFCs rendering

ehuss:
It seems like the markdown renderer used here doesn't allow starting a list next to a paragraph without a blank line. This prevents it from being rendered as a list.

I'm guessing there is a template or a script somewhere that generates this, since previous issues seem to have the same problem. You may want to update that.


nellshamrell (approved):
Ty!

PR #5771, opened 2024-07-30
Add Rust NYC, August 20th

colelawrence:
2024-08 Doing the Bare Minimum with Isograph - Rust NYC - Meetup Cover (Materialize)
https://www.meetup.com/rust-nyc/events/302480064/


nellshamrell (approved):
TY!

PR #5811, opened 2024-08-09
Announcing Aura 4

fosskers:
Aura's port from Haskell to Rust was recently completed. I've included links to Aura's User Guide that explain why the port was made in the first place.


cdmistman (approved):
thank you!

PR #5812, opened 2024-08-10
The weird of function-local types in Rust

Veetaha:


cdmistman (approved):
very fun and indeed weird! thank you!

PR #5819, opened 2024-08-14
Add Charlie Marsh interview

timClicks:
Charlie Marsh is the founder of Astral, the company behind Python’s fastest code formatter: ruff. Astral is also the home of the open source uv and rye projects, which make installing Python packages much faster than they have been with pip. All three are written in Rust. In the interview, Charlie discusses many things, including his own career, building an open source business backed by venture capital funding, growing a globally distributed team and operating successfully in the open.


cdmistman (approved):
awesome, thank you!

PR #5821, opened 2024-08-14
Adding information on supporting SBOM in Rust from Ferrous Systems

sabine-devins:
Updated with blog from Ferrous Systems on SBOM support in Rust and cargo in Tooling.


sabine-devins:
Apologies - did not realise that this had already been added to the newsletter.

PR #5822, opened 2024-08-14
Add "Fear Not the Association of Types," an article on using associated types in Rust

gavinleroy:


cdmistman (approved):
nice, thank you!

PR #5826, opened 2024-08-14
Add Zeeshan Ali Khan interview

timClicks:
Adds an interview from the Compose podcast. From the episode description:

> Zeeshan Ali Khan is the maintainer of the zbus crate, a pure Rust implementation of the D-Bus protocol. Zeeshan has been around Linux and embedded/mobile Linux for a long time – even moving to Finland to work for Nokia before the company moved to Windows Phone.
>
> The interview starts with creation of zbus—a pure Rust implementation of the D-Bus protocol that started on a whim—and how Rust can actually offer ease of use, as well as performance and memory safety.
>
> Expect lots of technical depth. We cover technical challenges like with using serde with binary file formats which are not self-describing, the use of unsafe code, and the future of Rust in Linux system agents. It turns out that the average kernel developer is not as fond of Cargo as the average Rust developer. Regardless, there is a huge opportunity — perhaps necessity — to bring memory safety into Linux. Should we rewrite of systemd in Rust?

### DIFF
--- a/draft/2024-08-14-this-week-in-rust.md
+++ b/draft/2024-08-14-this-week-in-rust.md
@@ -45,6 +45,8 @@ and just ask the editors to select the category.
 
 ### Miscellaneous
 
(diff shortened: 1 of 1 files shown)


cdmistman (approved):
thank you!

PR #5839, opened 2024-08-18
fix meetup group link

szabgab:


mariannegoldin (approved):
Thank you for catching and for the fix, approved

PR #5849, opened 2024-08-22
Added Rust Walkthrough: Using Diesel with custom array data types in Postgres

marvin-hansen:
This is a new end to end tutorial that was recently contributed to the Diesel crate.

While the PR is still under review and the update of the website is going to take some time too, 
it would be great if people can already access the repo from which the PR was derived since Postgres
 composite types  and custom arrays in Diesel can be quite tricky and this tutorial sheds some light on it.

### DIFF
--- a/draft/2024-08-21-this-week-in-rust.md
+++ b/draft/2024-08-21-this-week-in-rust.md
@@ -41,6 +41,8 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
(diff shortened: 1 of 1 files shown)


cdmistman:
thank you for your contribution! it's my understanding that you're waiting for the blog post to be released? if so we'd prefer to wait until that point to share the article - we don't want to share duplicate content 😄

PR #5850, opened 2024-08-23
publish 561

cdmistman:


jofas:
Did this work as intended? I can see [561 on the website](https://this-week-in-rust.org/blog/2024/08/21/this-week-in-rust-561/), but haven't received an email as is usually the case. Not sure how publishing TWiR works, but the file for 561 is still in the [`draft/`](https://github.com/rust-lang/this-week-in-rust/tree/master/draft) and not in the [`content/`](https://github.com/rust-lang/this-week-in-rust/tree/master/content) directory, which might be the reason no email was sent out to newsletter subscribers?

PR #5851, opened 2024-08-23
update draft template with the latest contents

cdmistman:


bennyvasquez:
@cdmistman I'm not sure I understand this update. Can you add some more context?

PR #5857, opened 2024-08-24
RFL-related LWN article

ojeda:
One more article from LWN that is freely available now.


bennyvasquez:
I'm including this, but please note that we normally omit articles that weren't published *this week*. Thanks for your submission!

ojeda:
Thanks!

Please note that LWN articles are not freely available when they are published, but a week after, which is why I send them after I get their notification of availability.

If that is not OK, then I will avoid submitting them.

PR #5861, opened 2024-08-27
add link to RustASIA conference

allenwyma:


bennyvasquez:
Hi there! Events have their own section in the newsletter and should be added following those protocols. Here's the relevant part of the display on every newsletter:

> If you are running a Rust event please add it to the [calendar](https://www.google.com/calendar/embed?src=apd9vmbc22egenmtu5l6c5jbfc%40group.calendar.google.com) to get it mentioned here. Please remember to add a link to the event too. Email the [Rust Community Team](mailto:community-team@rust-lang.org) for access.

allenwyma:
hi @bennyvasquez i missed this part, thanks for the information, i'll resubmit once i get it ready.

bennyvasquez:
Awesome! We'd love to have it.

allenwyma:
Hi @bennyvasquez the process isn't so clear to me how this works, but here's my understanding, please correct me if i'm wrong:

1. I have the event on the calendar, now. Will it be automatically added? Also the event is in Asia, so i'd like to have it published sooner, as it will probbaly need attendees to prepare more than 1 month ahead of time so they can arrange their schedule.

2. There's a current CFP process, should I just look back to previous episodes and follow how it's posted and submit a PR?

Updating the README with instructions on what we need to do to submit events would be helpful. I tried searching for some of the keywords from what you sent me above, but I cannot find what the process is.

PR #5863, opened 2024-08-27
Optimizing and measuring performance of zlib in rust

erikjee:
Blog post by @folkertdev, maintainer of zlib-rs. 
Covers current performance, SIMD optimization, measuring performance compared to c impl (zlib-ng).

Not sure what heading to put it under... Observations/Thoughts?

Thanks!

### DIFF
--- a/draft/2024-08-28-this-week-in-rust.md
+++ b/draft/2024-08-28-this-week-in-rust.md
@@ -38,6 +38,7 @@ and just ask the editors to select the category.
 ### Project/Tooling Updates
 
 ### Observations/Thoughts
+- [Current zlib-rs performance](https://trifectatech.org/blog/current-zlib-rs-performance/)
 
 ### Rust Walkthroughs
 


bennyvasquez:
Hi there! Thanks so much for this submission. To ensure the newsletter stays relevant and current, we limit submissions to those posts created in the past week, however. Feel free to submit articles as they are published in the future!

erikjee:
Ooh, that's too bad. I didn't know that. We published it Aug 8, but not shared it anywhere because we were waiting for a cross-post, here https://tweedegolf.nl/en/blog/134/current-zlib-rs-performance. (Aug 26)
That was well received on mastodon btw, see https://fosstodon.org/@tweedegolf/113027743789863839, so it would be interesting for more folks I guess. 
Would appreciate the article being included, if not we will be quicker in the future.
Thanks!

bennyvasquez:
Thanks for the additional context! In the future, it's best to either use the URL that points to the more recent of the posts, or submit before it's cross-posted, so it's still within the week. Thank you!

PR #5865, opened 2024-08-28
Research:  On the Impact of Memory Safety on Fast Network I/O

RiccardoSagramoni:
This paper attempts to critically evaluate the claims of high performance and memory safety associated with Rust, particularly in the context of low-level network programming. The approach involves rewriting Nethuns, a fast C-based network I/O library, using Rust.


bennyvasquez:
Thank you!

PR #5867, opened 2024-08-28
2024-08-28: Add Why Amazon, Cloduflare and Discord are building servers in Rust but you should probably not

sylvain101010:


bennyvasquez:
Thank you!

PR #5868, opened 2024-08-28
Move 561 into archive of past issues

mariannegoldin:


bennyvasquez:
We'll want the publish PR from last week's issue from @cdmistman, which should include this. If it doesn't, then I'll merge this one, too.

bennyvasquez:
fixed in 5870

PR #5885, opened 2024-09-04
Add backon to this week in rust

PsiACE:
BackON, a crate designed to make retrying a built-in feature in Rust, has now reached v1.


mariannegoldin (approved):
Thanks! Approved

PR #5888, opened 2024-09-04
2024-09-04 submission: "Why Rust needs scoped generics"

ais523:
A submission for the "Observations/Thoughts" section of This Week In Rust.

### DIFF
--- a/draft/2024-09-04-this-week-in-rust.md
+++ b/draft/2024-09-04-this-week-in-rust.md
@@ -39,6 +39,8 @@ and just ask the editors to select the category.
 
 ### Observations/Thoughts
 
+* [Why Rust needs scoped generics](http://ais523.me.uk/blog/scoped-generics.html)
+
 ### Rust Walkthroughs
 
 ### Research


mariannegoldin (approved):
TY! Approved

madsmtm:
Don't know where to discuss this, so thought I'd write quickly here:

Very insightful post! The problem of passing values to types reminds me of [contexts and capabilities](https://tmandry.gitlab.io/blog/posts/2021-12-21-context-capabilities/) that was discussed a few years ago - I suspect that "just" contexts and capabilities + generativity would be able to solve the problems that you outlined to begin with, though I am unsure, in any case your proposal is certainly more powerful!

ais523:
> Very insightful post! The problem of passing values to types reminds me of [contexts and capabilities](https://tmandry.gitlab.io/blog/posts/2021-12-21-context-capabilities/) that was discussed a few years ago - I suspect that "just" contexts and capabilities + generativity would be able to solve the problems that you outlined to begin with, though I am unsure, in any case your proposal is certainly more powerful!

Thanks for the feedback! (And thanks especially for the link – it's sent me down a rabbit hole of prior art on this subject.) I agree that this is effectively a cross between automatic context passing and generativity – but it's interesting that many of the comments on the automatic context passing proposals thought that it would need some sort of link to the type system to be useful.

PR #5892, opened 2024-09-05
Tiny typo fix in community item

obi1kenobi:
I believe that title component is intended to be "single mutex" not "singlemutex".


mariannegoldin (approved):
Thanks for catching! Approved.

PR #5901, opened 2024-09-09
Next-gen builder macro Bon 2.2 release 🎉. Derive syntax and cfg support 🚀

Veetaha:


mariannegoldin (approved):
Thanks so much, approved

PR #5908, opened 2024-09-10
Add Prossimo rav1d blog posts

crd477:


crd477:
This might be more appropriate for a different section of the page; I didn't think about it too hard.

mariannegoldin (approved):
Thanks, this is an interesting set of articles, always keen to see approaches for working with C code. Approved.

mariannegoldin:
Reconciled this with the existing set of observations/walkthrus.

PR #5910, opened 2024-09-11
Added an article about module-companion pattern

JohnScience:


mariannegoldin (approved):
Interesting approach, thanks for the PR! Approved.

PR #5913, opened 2024-09-11
Add a post on how checked queries in SQLx

matildasmeds:
This is a blog post that does an overview on the usefulness of checked queries in SQLx, and how to make any query checked. I was not sure about the correct section, so please feel free to move it as you see fit! 🙂


mariannegoldin (approved):
Thanks much! This works in the suggested section.

PR #5922, opened 2024-09-14
Next-gen builder macro Bon 2.3 release 🎉. Positional arguments in starting and finishing functions 🚀

Veetaha:
Hi! Just wanted to clarify, if you think I'm suggesting these blog posts to TWIR a little too often? I'll try to keep the updates batched in bigger blog posts and post them less often. But anyway, I'll keep this weekly update here this time (I've already published this update elsewhere on different platforms, so it's a bit too late to push back). I will reduce the updates cadance for the future posts though.


bennyvasquez:
One a week is just fine! Keep 'em coming. :) Thank you!

PR #5925, opened 2024-09-16
Add new article to "Rust Walkthroughs"

marioortizmanero:
This is an introductory article to the concept of "Subtyping and Variance". Let me know if the section is okay :)


bennyvasquez:
Thank you!

PR #5946, opened 2024-09-24
Extend rustc-perf triage report with days missing from yesterdays report

pnkfelix:
yesterday's triage report was missing a large chunk of days due to some website issue.

I don't know how to resolve the website issue itself, but I was able to manually work around the gap in time by hand.

This revision to the report hopefully deals with that.


nellshamrell (approved):
Ty!

PR #5951, opened 2024-09-25
fix: moved Rust Malaysia Meetup to be under Asia

shanipribadi:


mariannegoldin (approved):
Thank you so much for catching this! Approved

PR #5957, opened 2024-09-27
fixes create email script

nellshamrell:


andrewpollack:
Pushed changes to this PR to shift all publishing shebangs to be bash

andrewpollack (approved):

mariannegoldin (approved):
LGTM!

cdmistman:
I'm sorry I'm a little confused why this PR is open. What was broken with /bin/sh? /bin/sh should exist on all unix-compatible systems, but /bin/bash isn't guaranteed to do so (due to licensing). i understand we're using an ubuntu docker image but want to make sure i understand the problem :)

andrewpollack:
@nellshamrell was running into issues with using `source file.sh` over the WSL of Linux, and resolution was to update this shebang. I dont have the hardware to confirm that this was independent of the issue, and will need to defer explanation beyond this

nellshamrell:
it did not work on WSL, and when I changed it it did work, that is my best explanation :)

nellshamrell:
Just confirmed that is still the case on my system - this appears to be necessary for use with WSL

PR #5962, opened 2024-09-28
Rust for Linux LWN articles

ojeda:
A couple of articles (which became freely available recently) and some others for which I asked LWN for permission to use subscriber links.

### DIFF
--- a/draft/2024-10-02-this-week-in-rust.md
+++ b/draft/2024-10-02-this-week-in-rust.md
@@ -36,6 +36,16 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
+* [A new version of modversions](https://lwn.net/Articles/986892/)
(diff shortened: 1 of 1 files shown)


ojeda:
Added two more that got published since then.

ojeda:
Added one more from today.

ojeda:
One more I noticed in today's "newly freed" reminder.

cdmistman (commented):
wow, so many interesting articles, all very good reads :D 

however, I'm concerned about a 'flood' of lwn articles appearing this week 😅 I'll approve this, with the caveat that I might remove some articles depending on what the edition looks like. I also think a couple of articles are in the wrong section, so that will help :)

ojeda:
It is due to all the related conferences being together in September and the fact that I didn't have time to submit this every week while I was traveling -- it is not something expected every week.

If there is a limit of articles from a single source or something like that, we could put the older ones and next week newer ones, for instance.

PR #5965, opened 2024-09-30
Add "Sqlx4k - Interoperability between Kotlin and Rust, using FFI (Part 1)" under Rust walkthroughs.

smyrgeorge:
Add "Sqlx4k - Interoperability between Kotlin and Rust, using FFI (Part 1)" under Rust walkthroughs.


cdmistman (approved):
thank you!

PR #5969, opened 2024-10-01
added blog about rust interop with python and javascript

LisetRoos:
Blog post by Folkert de Vries about Rust interop in practice (with python and javascript).


cdmistman (approved):
awesome, thank you!

PR #5985, opened 2024-10-04
rust walkthroughs - Index Trait, Pinned Elements and Immutable Push V…

orxfun:
An article about a known limitation of the Index trait, and a workaround using pinned elements and immutable push vector.

[https://orxfun.github.io/orxfun-notes/#/imp-vec-motivation-2024-10-03](https://orxfun.github.io/orxfun-notes/#/imp-vec-motivation-2024-10-03)


cdmistman (approved):
nice, thank you!

PR #5990, opened 2024-10-07
microBioRust crate proposal and CFP projects

LCrossman:
Link to the explanation: 
[here](https://lcrossman.github.io/)
CFP links to issues under CFP section requests

Thanks for consideration


cdmistman (changes requested):
please make your changes to the dated file in the draft/ directory. i'll include these in a separate pr. thank you

PR #5991, opened 2024-10-08
added interview with Rust main architect at Volvo

LisetRoos:
A bit unsure whether I put it in the correct category.


cdmistman (approved):
awesome, thank you!

PR #5992, opened 2024-10-08
C/QotW and notable changes

llogiq:


YJDoc2:
Hey @llogiq , not directly related to changes in PR, but had a question regarding the `crate-of-the-week` suggestion. Let me know if there is another/better place to discuss this.

 Is there any known issues with why there have been no suggestions for cotw on rust-user-forum? Is it possible that people who might want to suggest a crate might not be on forum / want to join forum (for whatever reasons)? This will be an issue no matter what location (forum, github,reddit) is chosen, but is it possible that people on forum who might want to suggest a crate have already done so and we have reached a saturation point?

Another thing is from what I saw cotw was started in 2015 when Rust was not as widespread as it is now, so discoverability of crates was more of a problem. Now that there is a large and active community maybe it feels a big bar to reach for suggesting `crate of the week` - for me when reading this it feels like the suggested crate should be a "big" thing, which solves some important problem or does some complex thing. However actually if I'm not mistaken, it can be any crate. So maybe changing it to `community showcase` and changing the frequency to every other week or monthly might help in reducing the issue?

And thanks for creating and maintaining cotw till now, it helped me discovering some cool crates!

cdmistman (approved):
thank you!

llogiq:
@YJDoc2 The crate of the week was never to be The Big One, it was to celebrate the work of people that we might otherwise overlook, and give them the warm fuzzy feeling to have made it to CotW.

The rust-users discourse forum is the official channel for the Rust community. If people avoid that, so be it.

I will continue to select crates each week whether there are suggestions or not, and I will continue to name it Crate of the Week.

YJDoc2:
@llogiq , ok got it :+1: Thanks for the reply and doing cotw!

PR #6002, opened 2024-10-10
Add call for testing item

joshlf:
This links to https://github.com/google/zerocopy/discussions/1802. It's not an RFC like normal "call for testing" items, but hopefully it's in the spirit of this section.


joshlf:
Obsoleted by #6007

PR #6007, opened 2024-10-10
Add call for testing item

joshlf:
This links to https://github.com/google/zerocopy/discussions/1802. It's not an RFC like normal "call for testing" items, but hopefully it's in the spirit of this section.

*Obsoletes #6002*


bennyvasquez:
Hey @joshlf! This section has historically been for official RFCs, not for project RFCs. I'm going to decline this one, but have a question in to the normal editor of this section to ensure that is the correct choice. If not, you'll hear from us about how to correct it!

PR #6030, opened 2024-10-18
Correct 2024-10-16-this-week-in-rust.md to show title for article properly

jacks0n9:
The "Memory for Nothing" title said "Why Vec is (probably) a bad idea," however, it should say "Why Vec\<usize> is (probably) a bad idea. I assume this happened due to \<usize> being in the form of an html opening tag, and someone didn't escape it.


bennyvasquez:
Good catch! Thank you!

PR #6032, opened 2024-10-19
Propose tooling update for hifitime v 4.0.0

ChristopherRabotin:
Hifitime version 4.0.0 has been released. Hifitime is a precision oriented time management library in Rust. Version 3.0.0 was released just over two years ago.


bennyvasquez:
Thank you!

PR #6051, opened 2024-10-28
Dyn Box Vs. Generics

veeso:
I've added my new blog post: https://blog.veeso.dev/blog/en/dyn-box-vs-generics-in-rust/


nellshamrell (approved):
Ty!

PR #6060, opened 2024-10-30
Rust LWN articles

ojeda:

Thanks!


ojeda:
@nellshamrell This is the one I was referring to. Perhaps it should go to "project updates", instead of after the other LWN ones?

nellshamrell:
Yep! I will go ahead and move it there :)

PR #6063, opened 2024-10-31
C/QotW and notable changes

llogiq:
For some reason, this PR was lost when I did it the first time.


nellshamrell:
Will bring these into the published version!

nellshamrell:
Added in #6066

PR #6070, opened 2024-11-04
observation-thought: reached the recursion limit at build time

veeso:
Added this article from my blog: https://blog.veeso.dev/blog/en/reached-the-recursion-limit-at-build-time/


nellshamrell (approved):
Ty!

PR #6078, opened 2024-11-05
add post about the size of Futures

hds:
We explore why the size matters, how we can measure it, and how we can
check for futures that are very large with Clippy and Tokio Console.

Thank you for your work on This Week in Rust!


nellshamrell (approved):
Ty!

PR #6080, opened 2024-11-06
add entry for Select HTML Components in Declarative Rust

ifsheldon:
Hi! I'd like to add an entry for my blog post, Select HTML Components in Declarative Rust, which discuss the motivation, implementation and usage of my crate `scraper_query`. This blog has English and Chinese versions.


nellshamrell (approved):
Ty!

PR #6088, opened 2024-11-07
add m-z events in 572 to correct location

bdillo:
Fixing putting them in the wrong place from https://github.com/rust-lang/this-week-in-rust/pull/6087 :)


nellshamrell:
Ty!

PR #6090, opened 2024-11-07
Add Hyperlight entry

andreiltd:
Add Hyperlight project: https://opensource.microsoft.com/blog/2024/11/07/introducing-hyperlight-virtual-machine-based-security-for-functions-at-scale/


mariannegoldin (approved):
TY!

PR #6102, opened 2024-11-12
Rust LWN articles

ojeda:

Thanks!

### DIFF
--- a/draft/2024-11-13-this-week-in-rust.md
+++ b/draft/2024-11-13-this-week-in-rust.md
@@ -36,6 +36,8 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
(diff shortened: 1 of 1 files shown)


ojeda:
The first one is not a full article, but there was some discussion nevertheless about the blog post. Not sure what the editors' policy on those, so I am adding a note here. Thanks!

mariannegoldin (approved):
Thank you!

ojeda:
You're welcome!

ericseppanen:
In the future I don't think we should include links to articles that just link to something else. Our preference is to just link to the original source. There are many places hosting comments about articles; we don't generally provide additional links for that reason.

ojeda:
Thanks for the details! Yeah, that makes sense, and I suspected something like that could be the policy already, which is why I added the note above. At least in this case the author of the blog post wrote a comment in LWN, linking to another post, so hopefully someone found it valuable nevertheless.

PR #6118, opened 2024-11-18
blog.veeso.dev: You don't (*always*) need async

veeso:
Added this blog post https://blog.veeso.dev/blog/en/you-dont-always-need-async/


cdmistman (approved):
interesting, thank you!

PR #6121, opened 2024-11-19
rust walkthroughs - vector traits, polymorphic inputs, monomorphizati…

orxfun:
…on and the orx-v crate

An article that discusses the requirement for vector traits and motivation behind the orx-v crate that provides the abstraction to unify the vectors.

https://orxfun.github.io/orxfun-notes/#/v-for-vectors-2024-11-18


cdmistman (approved):
nice! thank you!

PR #6123, opened 2024-11-19
Blog post on WASM SIMD optimizations in zlib

erikjee:
Blog post on WASM SIMD optimizations used for zlib algorithms in zlib-rs and the results of benchmarking.


cdmistman (approved):
wow, neat! thank you!

PR #6132, opened 2024-11-20
community links for issue 574

ericseppanen:
I'm a bit tight on time today so I decided to take a shortcut and link to the RustConf and EuroRust youtube playlists instead of breaking out every individual video.


cdmistman (approved):
that's fine, i'll do it before publishing. thank you!

PR #6146, opened 2024-11-27
updating events for twir-575

bdillo:
Some explanation for some of the deleted entries that are still within the date range:
- 2024-11-28 Hacker Dojo meetup was cancelled
- 2024-12-07 Rust Circle Kampala was listed twice, once under `Virtual` and once under `Africa`, kept the `Virtual` one


mariannegoldin (approved):
TY! Approved with some updates.

PR #6167, opened 2024-12-04
C/QotW and notable changes

llogiq:


cdmistman (approved):
can confirm 'self own' is a rust thing, i do it all the time

PR #6175, opened 2024-12-05
Add "Breakage! in the Cargo.toml" post

obi1kenobi:
It's a feature highlight post for the newly-released `cargo-semver-checks v0.37`, which is now able to check package manifests for breakage.


cdmistman (approved):
Wow very cool! thank you!

PR #6177, opened 2024-12-06
remove duplicate entry

szabgab:
Apparently it was also added with some automated tool


bennyvasquez:
Good catch. Thank you!

PR #6197, opened 2024-12-14
Add Hex Display announcement

parasyte:
I recently announced a new small CLI tool that is comparable to `xxd`. It's called Hex Display or `hd`, and it has some nice features.


bennyvasquez:
Thank you!

PR #6208, opened 2024-12-18
Add a link to Fjädra, a force-based graph layout library

grtlr:
Fjädra (or `fjadra`) is a force-based layout engine written in pure Rust without dependencies. It aims to port the very popular [d3-force](https://d3js.org/d3-force) from JavaScript as closely as possible.

### Disclaimer

This library powers the graph view of Rerun which is mentioned here: https://github.com/rust-lang/this-week-in-rust/pull/6207/files. However, Fjädra is standalone and does not require Rerun—I still thought it may be worth mentioning.

If people think that these two releases are too closely related, please prefer the PR linked above. Thank you.

### DIFF
--- a/draft/2024-12-18-this-week-in-rust.md
+++ b/draft/2024-12-18-this-week-in-rust.md
@@ -37,6 +37,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


bennyvasquez:
Thank you!

PR #6211, opened 2024-12-19
Add "So many tokens, so little time"

look:
This is a blog post about a new byte-pair tokenizer written in Rust.

The source for the crates themselves is in the [bpe](https://github.com/github/rust-gems/tree/main/crates/bpe) and [bpe-openai](https://github.com/github/rust-gems/tree/main/crates/bpe-openai) directories.


bennyvasquez:
Thank you!

PR #6223, opened 2024-12-21
Meetup group updates

bdillo:
I went through the groups in `meetups.json` and split this file into `rust-meetups.json` and `maybe-rust-meetups.json` for better use in our meetup scripts (still pending). `rust-meetups.json` contains rust-specific meetup groups, `maybe-rust-meetups.json` contains groups that occasionally have rust events, so these events must be manually looked at by an editor.

Some criteria for where the groups land in each file and why some are removed from the previous `meetups.json`:
- If a group no longer exists, remove it
- If a group still exists and is a rust specific group, keep it in `rust-meetups.json`, even if it has been inactive (no real overhead from checking, might as well leave them here in case they become active again)
- If a group isn't rust specific and hasn't had a rust-specific event within the last year, remove it. If it has had a rust event in the last year, put it in `maybe-rust-meetups.json`


nellshamrell (approved):
Awesome! Ty!

PR #6246, opened 2025-01-01
Removing 2 events

szabgab:
one was a Python event the other one was Rust, but it was unfortunately cancelled


mariannegoldin (approved):
Thanks for the catch on the Python event, and the update on the cancelled event. Approving.

PR #6305, opened 2025-01-20
add "Introducing RealtimeSanitizer for Rust"

steckes:
This blogpost announces the release of `rtsan-standalone` and discusses how the sanitizer that is a path of the upcoming LLVM version could be integrated directly into Rust.


nellshamrell:
Looks like we got this in another pull request, but ty for making sure we saw this!

PR #6323, opened 2025-01-24
Add Rust Bern meetup event

dspicher:


bdillo:
Hey, thanks for the PR to add this event. For meetup groups, we have a script that pulls events from a list of known rust groups, (including this group) https://github.com/rust-lang/this-week-in-rust/blob/f5e255de4148b07e8c49a9ce97f6955ee2f7670e/tools/events/rust-meetups.json#L53

Each week we'll get the new events from the groups in the list when we run our scripts, so no need to add these manually. I just ran the script for this week and the event is included there (#6329). Again thanks for the PR, if you have any other meetup groups feel free to submit a PR to add them to the list and let us know if you have any questions.

dspicher:
Oh wow, what a fantastic change, thanks for the explanation @bdillo!

PR #6337, opened 2025-01-31
add vodei "Are we embedded yet?"

szabgab:
Implementing tiny HTTP server on a microcontroller using Rust with Maor Malka


bennyvasquez:
Thank you!

PR #6339, opened 2025-02-02
Meetup json output, minor fixes

bdillo:
This fixes some of the event location capitalization issues. The meetup API returns strings like "San francisco", so this fix will format these things a bit more correctly. Also in GB part of the postal code was used as the state, this is ignored now. Additionally, the meetup API returns an empty string if a part of a location is missing, which we now switch to `None`.

This also adds a way to output the events as json so they can be used in further automation. Currently we're just merging the lists together manually, having a better output format to use will let us create other tools to merge the lists together and save us some time.

Change was tested manually, example json partial output:
```
{"Virtual": [{"name": "Indy.rs - with Social Distancing", "location": "Indianapolis, IN, US", "date": "2025-02-05", "url": "https://www.meetup.com/indyrs/events/302031658", "virtual": true, "organizer_name": "Indy Rust", "organizer_url": "https://www.meetup.com/indyrs/events/"}, ...
```

### DIFF
--- a/tools/events/meetup-automation/event.py
+++ b/tools/events/meetup-automation/event.py
@@ -82,6 +82,18 @@ class Event:
     self.name = self.name.strip()
     self.organizer_name = self.organizer_name.strip()
 
+  def to_dict(self) -> dict:
+    """ Method for serializing to a dict which can be further serialized to json """
+    return {
+      "name": self.name,
+      "location": self.location.to_str(),
+      "date": self.date.strftime("%Y-%m-%d"),
+      "url": self.url,
+      "virtual": self.virtual,
+      "organizer_name": self.organizer_name,
+      "organizer_url": self.organizer_url
+    }
+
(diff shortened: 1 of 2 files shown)


bennyvasquez:
I'm going through PRs in prep for this week's edition. I'm not at all qualified to review this one (which is why you didn't ask me to!) but I'm excited for it nonetheless! We'll leave it for @mariannegoldin. :D

mariannegoldin (approved):
Thanks very much for this time saving improvement!

bdillo:
> Hi @bdillo - manually tested, and works great.
> 
> Is the intention to run the script over `maybe-rust-meetups.json` and `rust-meetups.json` and then manually collate the results?

for now yes, I'll probably just add a basic check to see if the string "rust" (case insensitive) shows up in the maybe-rust meetup groups

PR #6366, opened 2025-02-12
Modeling Inheritance in SeaORM

tyt2y3:
* [Modeling Inheritance in SeaORM](https://www.sea-ql.org/blog/2025-01-08-sea-orm-inheritance/)


bennyvasquez:
Thank you!

PR #6393, opened 2025-02-19
Scaling PayPay with Rust

ranger-ross:
The good folks at PayPay recently started using Rust and they published a blog post with their experience. 

The article is originally in English but was translated into Japanese by the translation team.


nellshamrell (approved):
Ty!

PR #6396, opened 2025-02-19
add an entry for Slack Your REST Client with a couple of Serde Tricks

ifsheldon:
Hi! This blog announces the new version of `async-openai-wasm` crate and details the learning I got when developing this new version. Most of it is about the learning, not the crate. So if you like, you can just move it to Rust Walkthroughs section. Thanks!


nellshamrell (approved):
Ty!

PR #6398, opened 2025-02-19
Add entry to "A length-indexed Vector in Rust"

guilleme-90:
I stumbled on this post about a technique to construct a kind of vector whose size is a compile-time constant.


nellshamrell (approved):
Ty!

PR #6404, opened 2025-02-20
fix(587): Fix broken link due to brackets

futile:
The URL contains brackets, i.e., `()`, which seem to throw off the Markdown-to-HTML rendering, leading to the link being broken.

This commit simply URL-encodes the brackets, resulting in a link that works and is not ambiguous w.r.t. Markdown-to-HTML rendering.


extua:
This would close #6404.

futile:
s/brackets/parentheses/g it would seem.. 😅

nellshamrell (approved):
Good catch! Ty!

PR #6407, opened 2025-02-22
Fold list of merged pull requests

Kobzol:
It is quite long and noisy, and IMO not *that* interesting to most readers. This was suggested [here](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/Ideas.20for.20improving.20the.20TWiR.20newsletter).

### DIFF
--- a/draft/DRAFT_TEMPLATE
+++ b/draft/DRAFT_TEMPLATE
@@ -106,11 +106,16 @@ If you are an event organizer hoping to expand the reach of your event, please s
 
 ## Updates from the Rust Project
 
-<!-- Rust updates go here -->
-
 ### Rust Compiler Performance Triage
 
 <!-- Perf results go here -->
+
+<details>
+<summary>Merged pull requests in the Rust toolchain</summary>
+
+<!-- Rust updates go here -->
+
+</details>
 
 ### [Approved RFCs](https://github.com/rust-lang/rfcs/commits/master)
 


draft/DRAFT_TEMPLATE
```diff
@@ -104,13 +104,18 @@ Are you a new or experienced speaker looking for a place to share something cool
 
 If you are an event organizer hoping to expand the reach of your event, please submit a link to the website through a [PR to TWiR](https://github.com/rust-lang/this-week-in-rust) or by reaching out on [X (formerly Twitter)](https://x.com/ThisWeekInRust) or [Mastodon](https://mastodon.social/@thisweekinrust)!
 
+### Rust Compiler Performance Triage
```
kpreid, line 107: Moving this up without any other changes makes it a child of the Call for Participation heading.

Kobzol: I missed the difference between `##` and `###`, it was confusing to me that "Updates from the Rust Project" just included a list of PRs :man_shrugging: Thanks!

nellshamrell:
Hello there! @llogiq has made some changes along these lines. I will let @llogiq reply here when able.

llogiq:
I shortened the list of merged pull requests instead. Let's try that, and we can fold it later if it's still too long.

llogiq:
The number of PRs already links to the underlying search. There is quite a lot of noise in there, though.

Kobzol:
Cool, didn't realize that.

PR #6422, opened 2025-03-01
Slight wording tweak in 2/26

scottmcm:
`slice::Iter::next` was always inlined in LLVM; the piece that's new is that it inlines *in MIR*.

I propose this slight tweak to avoid anyone thinking "wait, it didn't inline before?!?" when reading this, because slice iterators not inlining would have meant horrible runtime performance.

cc @rylev


rylev (approved):

bdillo:
Obviously this one has already been sent out, but thank you for the correction!

PR #6423, opened 2025-03-01
Update 2025-03-05-this-week-in-rust.md

ifdario:


bdillo:
Thank you for the contribution, but unfortunately we are not accepting just plain links to a github repo. If you write up a post with some more detail we would be happy to include that. Nice readme though, lots of detail!

PR #6429, opened 2025-03-04
Adding Rust Trends Newsletter and Post (www.rust-trends.com)

Rust-Trends:
Adding Rust Trends Newsletter and Post (www.rust-trends.com)


bdillo:
thanks!

PR #6447, opened 2025-03-07
add "Ratatui - terminal user interfaces in Rust with Orhun Parmaksız - build ratatop in pair programming"

szabgab:


bdillo (approved):
thanks!

PR #6449, opened 2025-03-08
Add "when is it public API for a trait to be implementable" blog post

obi1kenobi:
It describes a number of edge cases around trait sealing that I discovered and had to reason through & implement in cargo-semver-checks. It includes the headline features of cargo-semver-checks v0.40.


bdillo (approved):
read this the other day, interesting stuff, congrats on the new release as well!

obi1kenobi:
Thank you, I'm glad you found it interesting! 😁

PR #6450, opened 2025-03-09
Add Build A Web + Desktop Application With Rust walkthrough

Kofituo:
I've added the links now

### DIFF
--- a/draft/2025-03-12-this-week-in-rust.md
+++ b/draft/2025-03-12-this-week-in-rust.md
@@ -41,6 +41,8 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
+* [Create A Web + Desktop Application With Rust](https://medium.com/@otukof/create-a-web-desktop-application-with-rust-c8449f661ecc)
+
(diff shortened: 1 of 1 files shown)


bdillo:
hey, thanks for the submission again, unfortunately I don't see any actual rust content in the link, I think we'd like to see a little bit of rust code at least. If you want to see some examples of the kind of content we usually accept you can check out the other submissions in the newsletter. We also have some guidelines here https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#what-do-we-look-for-when-considering-whether-to-include-something-in-this-week-in-rust

Kofituo:
Hello,
It's an article series. That article is an introduction. Within the article, I provide links for the subsequent parts. I was thinking it'll be best to submit the article series, starting from the introduction.

bdillo:
I think the criteria here is that when you open a link, there is a good amount of detailed rust content. We don't want people to have to follow other links, just straight to the content is what we want.

Kofituo:
Okay sure.

PR #6455, opened 2025-03-11
add meetup announcement for Rust Rhein Main in March

christophbeberweil:
Hi everyone, 

please add our latest meetup announcement.

Thanks,

Christoph


bdillo (approved):
thanks! also we should be pulling these in from some of our events automation scripts, looks like we have the meetup group here: https://github.com/rust-lang/this-week-in-rust/blob/master/tools/events/rust-meetups.json#L92, so you generally should not need to submit these!

PR #6460, opened 2025-03-11
async-rust-for-dummies

veeso:
Added this article to walktroughs

https://blog.veeso.dev/blog/en/async-rust-for-dummies/


bdillo (approved):
thanks!

PR #6463, opened 2025-03-12
Add AIScript

Folyd:
https://aiscript.dev


bdillo:
hey, thanks for the submission, unfortunately we do not accept plain links to github repos (see https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#projectstooling-updates) . Happy to include a different link if it meets the criteria.

Folyd:
Can I use [https://aiscript.dev](https://aiscript.dev/) as the link?

bdillo:
I think something like https://aiscript.dev/blog/announcing-aiscript would fit a little better, ideally if you could add a bit more rust content in it too, that would be great!

PR #6479, opened 2025-03-18
Add blogpost about algorithm implementation that is faster in Rust due to defined casting semantics

AdamGS:
Pretty cool find we ran into, including a detailed comparison with the cpp implementation.


mariannegoldin (approved):
Fascinating, Thanks for submitting!

PR #6492, opened 2025-03-20
vendoring c/cpp dependencies in rust

veeso:
added https://blog.veeso.dev/blog/en/vendoring-c-cpp-dependencies-in-rust/


mariannegoldin (approved):
Thanks, this is a great topic.

PR #6496, opened 2025-03-23
add "Safe Delayed Initialization for Lifetime Extension"

lolbinarycat:
add https://paper.wf/binarycat/safe-delayed-initialization-for-lifetime-extension


mariannegoldin (approved):
A subtle but important nuance, always nice to address this. Thanks!

PR #6498, opened 2025-03-25
Add a "Just write a test for it" blog post

Kobzol:


mariannegoldin (approved):
TY! I had to remember to collapse the example Rust code. I wonder (non-blocking), is there a way to have it open by default for ease of access? Approved to publish.

Kobzol:
I guess it could be possible with a bit of JavaScript. How should I detect when it should be opened though, is there some accessibility optiom for that? :thinking: Other people were happy that it was collapsed, so that it didn't interfere with the rest of the text.

PR #6504, opened 2025-03-26
feat: add cot v0.2 announcement

seqre:


mariannegoldin (approved):
Thanks for the submission and congrats on the release! In the future, we love to see examples of how was Rust was used in building tools, among other types of Rust-specific content, in the body of release posts. Our readership really appreciates it.

seqre:
That's a fair point, we'll make sure to include some examples next time, thanks for the tip!

PR #6510, opened 2025-03-27
Link to tutorial blog post "Building a CoAP application on Ariel OS"

chrysn:
If self-suggestions are allowed: I published an entry level tutorial on embedded development using CoAP and Ariel OS. It aims to pick up users who have little prior experience with Rust, so I hope the category fits.


mariannegoldin (approved):
Thanks so much for your contribution. Yes, we love self-directed submissions. Enjoyable article, with useful explanations. Consider submitting to the (unaffiliated) TER Embedded Rust newsletter as well.

PR #6514, opened 2025-03-30
Add "Public mdBooks"

szabgab:
Alternatively you might prefer to include a link to a post I made about it: https://szabgab.com/public-mdbooks

### DIFF
--- a/draft/2025-04-02-this-week-in-rust.md
+++ b/draft/2025-04-02-this-week-in-rust.md
@@ -44,7 +44,8 @@ and just ask the editors to select the category.
 ### Research
 
 ### Miscellaneous
-
+* [Public mdBooks](https://mdbooks.code-maven.com/)
+ 
 ## Crate of the Week
 
 <!-- COTW goes here -->


draft/2025-04-02-this-week-in-rust.md
```diff
@@ -44,7 +44,8 @@ and just ask the editors to select the category.
 ### Research
 
 ### Miscellaneous
-
+* [Public mdBooks](https://mdbooks.code-maven.com/)
```
mariannegoldin, line 47:
```suggestion
* [Public Rust mdBooks](https://mdbooks.code-maven.com/)
```

mariannegoldin: This addition might make it a little more clear that these are all Rust-related

szabgab: Except that not all of them are Rust related. Most of them are, but the point of the site is to collect all the (public) users of `mdbooks` which is a Rust project.

mariannegoldin: Hi @szabgab thanks for the clarification. It sounds like this is collecting users of the mdBooks Rust project, so it makes sense as a project update story.

mariannegoldin: This can stay in Misc as-is, or be moved to Project updates, up to you.

PR #6520, opened 2025-04-01
Embedding shared objects in Rust

veeso:
Added https://blog.veeso.dev/blog/en/embedding-shared-objects-in-rust/


mariannegoldin (approved):
Thank you!

PR #6524, opened 2025-04-01
add: SALT

mojeanmac:
Introduced a while back in #4920 and is a continuation of this research project, now including analysis for functional and imperatively-written code.


mariannegoldin (approved):
Thank you for the submission, and thanks for including information in the app description about how to opt-in to data gathering, and the plugin data gathering consent process. I will go ahead and approve this, with the non-blocking comment that it would be useful to include a link to the HSRB page for the project from the plugin description page, as the study is gathering individual data.

PR #6534, opened 2025-04-07
Add article to TWIR draft

trevoranderson:
Adds an article I wrote about how Serendipity Systems LLC uses Rust as the primary language business logic language. Is this in the right section? It might fit better under walkthroughs section, but I'll defer to maintainers on which section is best.


mariannegoldin (approved):
Nice explanation of the FFI case with a UI library, thanks for your submission! The observations section is a good place for the article.

mariannegoldin (approved):
Nice explanation of the FFI case with a UI library, thanks for your submission! The observations section is a good place for the article.

PR #6540, opened 2025-04-09
update social media links

ericseppanen:
Potential issues:
- I'm not sure what the canonical bsky link should be.
- Which order should links be?


mariannegoldin (approved):
Thank you very much! The format looks correct to me.

PR #6555, opened 2025-04-15
add rama content to 2025-04-16-this-week-in-rust.md

GlenDC:


nellshamrell (commented):
Hello there!

To include something in the Project/Tooling updates, we need a little more information in the release notes (for example, what are the changes, how do they apply to users, how do they use them, etc.)

However, the call for participation links look great and I will include those.

PR #6558, opened 2025-04-16
added blog about SIMD in zlib-rs

LisetRoos:
My colleage Folkert de Vries is writing a blog series explaining SIMD, using examples from zlib-rs. This blog, part 1, was published yesterday.


nellshamrell:
Looks like we got this in another pull request, but ty for submitting it! It is in the newsletter :)

PR #6565, opened 2025-04-18
Using a zig allocator in Rust

crustly:
[Zig -> allocators -> Rust ergonomics](https://www.capturedlambda.dev/blog/zig-allocators-rust_ergo/)

Not sure if this is in the right category.


nellshamrell (approved):
That category works great! Ty!

PR #6571, opened 2025-04-22
Updated RustWeek. Added RustNL hackathon

erikjee:
The RustWeek entry did not have the correct (best possible) links.

Also added the free hackathon RustWeek is hosting on May 16.


mariannegoldin (approved):
Thank you very much for the correction!

PR #6575, opened 2025-04-23
added blog about safer software with Rust

LisetRoos:
Adds blog 'Does using Rust really make your software safer? by Marc Schoolderman


nellshamrell:
Looks like we go this one in another pull request, but thank you for making sure we saw it!

PR #6577, opened 2025-04-23
Adding How To Get A Rust Job- Survey For Those Who've Managed To Get One

drewcapener:


nellshamrell:
Could you add some info to the form about who is collecting this info, how it will be used, etc? People may feel better about answering the survey if they know that.

nellshamrell:
Feel free to re-submit when you add that info!

PR #6569, opened 2025-04-22
Add perf triage for 2024-04-22

panstromek:

PR #6584, opened 2025-04-25
fix year in perf triage link in 2025-04-23

panstromek:
I accidentaly used 2024 year in name. see: https://github.com/rust-lang/this-week-in-rust/pull/6569#issuecomment-2830849524

The original link still works and points to the right doc, the new one is just less confusing.


bdillo (approved):
thanks for the fix!

PR #6585, opened 2025-04-26
Add file-editor 0.1.0 to Project/Tooling Updates

davidwilliam:
Adds file-editor crate to the Project/Tooling Updates section.
Provides a fluent API for creating, inserting, searching and replacing text in UTF-8 files.


bdillo:
thanks for the contribution, unfortunately we do not accept plain crate/repo links in the project/tooling updates section, see our policy: https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#projectstooling-updates

if you were to do a little writeup of your crate we would be happy to include it though!

PR #6588, opened 2025-04-27
Update participating meetups for anniversary meetup

jbethune:
This is a joint meetup of  10 local Rust meetups to celebrate 10 years of Rust 1.0.

It would be really sweet if we could get the names of all the participating meetups on the this-week-in-Rust newsletter. This way, it is easier for regulars of the individual meetups to find this joint meetup on the list. And other people might also get curious.

This PR is done is agreement with the organisers of the 10 meetups.

Please let me know if I should use a different formatting.

Thanks!


bdillo (approved):
wow this is great, thanks. we do keep track of the meetup groups we pull from with scripts each week here if you wanted to take a look: https://github.com/rust-lang/this-week-in-rust/blob/master/tools/events/rust-meetups.json

I'd like to make this documented a little better in the future though so people can add their own meetup groups more easily, but for now they live in that file.

PR #6589, opened 2025-04-28
Add new article - How Rolldown Works: Module Loading, Dependency Graphs, and Optimization Explained

ShroXd:
Add an article to Rust Walkthroughs section, this article explains the principle of the scan stage of Rolldown.
The link and the title of the article don't exactly match because my blog site wants to keep the link to the article concise.

I'm not sure if I put the article in the appropriate section, appreciate for any comments or editing.

### DIFF
--- a/draft/2025-04-30-this-week-in-rust.md
+++ b/draft/2025-04-30-this-week-in-rust.md
@@ -44,6 +44,8 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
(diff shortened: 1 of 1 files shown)


bdillo (approved):
I moved the article to the misc section as I don't think it quite fits under walkthroughs. you can see some our guidelines for categories here: https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#community-sub-categories

thank you for the contribution!

PR #6602, opened 2025-05-01
Fix walkthrough incorrectly marked as "[video]" in 2025-04-30-this-week-in-rust.md

Minimal-C:
Super minor fix: I believe this walkthrough article ([Generating 1 Million PDFs in 10 Minutes](https://www.ersteiger.com/posts/rendering-one-million-pdfs/)) was incorrectly marked as video when it is text based.


bdillo (approved):
whoops sorry about that, thanks for the fix.

PR #6613, opened 2025-05-06
added Rust Augsburg Meetup

tiagomanczak:
I organize the Rust Augsburg Meetup and would like to add the next edition to the Rust Week.


bdillo (approved):
thanks!

PR #6624, opened 2025-05-08
Link event about embedded Rust in Ghent

wvhulle:


mariannegoldin (approved):
Thank you! Approved with an ordering change.

PR #6625, opened 2025-05-09
Fix quoting and broken link in 2025-05-07-this-week-in-rust.md

teor2345:
One of the links in this week's tracking issues list has a quote outside the link bracket, which disables the link and breaks formatting.


ericseppanen:
Thank you!

PR #6640, opened 2025-05-15
Events 599

bdillo:
Events for 599, also updated some of the automation to not put `/events` at the end of the meetup URLs. The meetup URL shows the group description as well as upcoming events, so it seems a bit nicer.


mariannegoldin (approved):
Ty very much. The elision of /events is a nice touch.

PR #6641, opened 2025-05-15
Meetup script updates

bdillo:
Changes to the meetup scripts:
- UK events show UK as the country code in the meetup ui, but return GB. Normalize them to use UK as this seems like the more common way to write the addresses there
- Add support for location overrides. Some meetup groups are hybrid (like Seattle and Vancouver), so we can mark them as such in our rust-meetups file. Additionally, some meetup groups don't list their location at all, so we can mark those as virtual if needed, after manually checking they are in fact virtual.
- With the above changes, the rust-meetups file has a new format so we can add arbitrary properties for each meetup group if we want to have some specific behavior within our scripts.
- Also remove the `/events` part of the meetup group URLs. The URL without the `/events` section lists the groups description as well as the upcoming events, whereas `/events` just lists the events. Just seems like a nicer URL to use.

I manually tested and validated the data, also used this script as part of the process in #6640


tools/events/rust-meetups.json
```diff
@@ -1,124 +1,117 @@
...
+    "https://www.meetup.com/charlottesville-rust-meetup": {},
+    "https://www.meetup.com/chicago-rust-meetup": {},
+    "https://www.meetup.com/christchurch-rustlang-meetup-group": {},
+    "https://www.meetup.com/columbus-rs": {},
+    "https://www.meetup.com/copenhagen-rust-community": {},
+    "https://www.meetup.com/dallasrust": {},
+    "https://www.meetup.com/desert-rustaceans": {},
+    "https://www.meetup.com/detroitrust": {},
+    "https://www.meetup.com/dutch-rust-meetup": {},
+    "https://www.meetup.com/indyrs": {},
```
mariannegoldin, line 20:
```suggestion
    "https://www.meetup.com/ideas-and-coffee/": {},
    "https://www.meetup.com/indyrs": {},
```

bdillo: I think this group might go better into the "maybe rust" group list as all the events are not rust specific. I was actually going to fold that list into this one and add a field into this file to tell the script to just look for the string "rust" in the event names, so that way we don't have to manually filter through these.

PR #6642, opened 2025-05-15
added the joint/hybrid meetup of Rust Dortmund and the C++ and Rust Dragons

DarthB:
(former Rust Cardiff)

focused on Fullstack with Rust with talks by Ben Wishovich and Goetz Markgraf


mariannegoldin (approved):
Thank you!

PR #6645, opened 2025-05-16
Post: Joining threads immediately as they finish in Rust

jpmelos:
This is a post about joining multiple threads in Rust in a way that they can be joined immediately as they finish.

### DIFF
--- a/draft/2025-05-14-this-week-in-rust.md
+++ b/draft/2025-05-14-this-week-in-rust.md
@@ -44,7 +44,7 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
-- [Joining threads immediately as they finish in Rust](https://jpmelos.com/posts/joining-multiple-rust-threads/)
+* [Joining threads immediately as they finish in Rust](https://jpmelos.com/posts/joining-multiple-rust-threads/)
 
 ### Research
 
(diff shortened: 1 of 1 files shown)


draft/2025-05-14-this-week-in-rust.md
```diff
@@ -44,6 +44,8 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
+- [Joining threads immediately as they finish in Rust](https://jpmelos.com/posts/joining-multiple-rust-threads/)
```
mariannegoldin, line 47:
```suggestion
* [Joining threads immediately as they finish in Rust](https://jpmelos.com/posts/joining-multiple-rust-threads/)
```

mariannegoldin (approved):
Thank you! Approved with a syntax correction.

ericseppanen:
Thanks for the submission!

I was reviewing this post with one of the rust standard library authors, and we both felt that this article could use some improvement, so that it doesn't cause confusion among readers new to Rust. Specifically, it should mention the fact that `thread::spawn` can be used in "detached" mode (by just dropping the `JoinHandle`), where the caller does not need to call `join()` at all.

If we've misunderstood, feel free to continue the conversation. I'm going to revert the PR in this week's issue because we're already late and need to get it published soon, but feel free to resubmit again next week! Apologies for the mixed messages; it was a miscommunication on our end.

If a timely result is needed from the spawned thread, it's fairly common to pass a `Sender` to into the spawned function and have it transmit information that way, instead of using the return value and doing more work to loop over the `JoinHandle`s repeatedly. It might also help to mention this approach.

jpmelos:
Thanks, I'll edit the post and try again.

PR #6663, opened 2025-05-21
2025-05-21: Add Secrets managers considered harmful. How to securely encrypt your sensitive data with envelope encryption and KMS in Rust

sylvain101010:


bennyvasquez:
Hi there! This one is interesting, but doesn't actually include much rust-specific content, so I'm going to decline this. Thanks for your submission, and let me know if you have any questions!

PR #6666, opened 2025-05-22
Post: Joining threads immediately as they finish in Rust

jpmelos:
I tried to publish this article in the newsletter last week: https://github.com/rust-lang/this-week-in-rust/pull/6645.

@ericseppanen requested some changes [here](https://github.com/rust-lang/this-week-in-rust/pull/6645#issuecomment-2886972685), and I addressed those requests, so I'd like to try again.


bennyvasquez:
Thank you!

PR #6713, opened 2025-06-05
Add talk recording link

ianthetechie:
This is a talk I gave in January at the Seoul iOS Meetup on how (and why) to write your core business logic in Rust and link it as a shared library, with ergonomic bindings generated by UniFFI.


nellshamrell (approved):
Ty!

PR #6721, opened 2025-06-08
security: Run zizmor on CI, resolve actions

andrewpollack:
[zizmor](https://github.com/zizmorcore/zizmor) runs static analysis on GitHub actions with proposed remediation steps. I ran on the TWiR repo, and attached remediations for each:
* https://docs.zizmor.sh/audits/#artipacked `actions/checkout` will by default persist git configuration for the duration of the workflow, which is not necessary in this case.
* https://docs.zizmor.sh/audits/#excessive-permissions detects excessive permissions in workflows, both at the workflow level and individual job levels.
* https://docs.zizmor.sh/audits/#unpinned-uses Tag- or branch-pinned actions can be changed by the upstream repository, either by force-pushing over the tag or updating the branch.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -1,5 +1,9 @@
 name: Correctness Checks
 on: [push, pull_request]
+
+permissions:
+  contents: read
+
 jobs:
   Run-Markdown-Checks:
     runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


nellshamrell (approved):
Thank you so much @andrewpollack!

PR #6722, opened 2025-06-09
Add `Why doesn't Rust care more about compiler performance?` blog post

Kobzol:
https://kobzol.github.io/rust/rustc/2025/06/09/why-doesnt-rust-care-more-about-compiler-performance.html


nellshamrell (approved):
Ty!

nellshamrell:

PR #6740, opened 2025-06-16
Add Rust Compiler Performance Survey 2025

Kobzol:
https://blog.rust-lang.org/2025/06/16/rust-compiler-performance-survey-2025/


nellshamrell (approved):
Ty!

PR #6770, opened 2025-06-26
Rust LWN articles (Asterinas)

ojeda:
Asterinas was covered earlier this month by linking to arXiv, but there has been an LWN article about it some days which you may also want, so here it is too.

Thanks!


bdillo (approved):
thanks!

PR #6773, opened 2025-06-27
New project announcement: kiorg

houqp:


bdillo:
thank you for the contribution, however we do not accept plain links to github repos/crates per our [project/tooling guidelines](https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#projectstooling-updates), this is purely a volume thing, if we accepted things like this and links to crates we would simply have too many. If you made a little writeup as a blog post or something we would be happy to include that!

PR #6785, opened 2025-07-02
added augsburg meetup #14

tiagomanczak:
added our Augsburg meetup in Juli


mariannegoldin (commented):
Hi there, this fall outside of our range for the TWIR 606 issue (2 July - 30 July) but we will be sure to include it for next week!

PR #6789, opened 2025-07-05
Deep dive into temporaries

conectado:
Hi,

I wrote an article about temporaries in Rust and I'd like it if it could be included in next week issue. Thanks!


mariannegoldin (approved):
Very interesting topic, thank you.

PR #6817, opened 2025-07-15
Update 2025-07-16-this-week-in-rust.md

beryl678:
docs: recommend a rust blog


mariannegoldin:
HI there, I will have to pass on this submission, as it does not include sufficient Rust-specific content beyond the fact that the product was built with Rust. We prioritize posts with Rust-specific details for our readership. Thanks for the submission however!

PR #6862, opened 2025-08-02
610: Fix Markdown error

tbu-:


tbu-:
I feel like this is an error that could have been found by proofreading the whole issue once. (It's very visible because the link isn't in the rendered markdown, but rather just the URL). Is that not something that happens?

nellshamrell:
Good catch! While we do our best to avoid errors when publishing, sometimes something slips through. We greatly appreciate people opening pull requests to fix these errors like this, ty!

PR #6882, opened 2025-08-09
Add [video] tag for youtube.com links in 2025-08-06 file and add guidance

homersimpsons:
Should I also update the guidance in https://github.com/homersimpsons/this-week-in-rust/blob/c8eeb63e2b83a99109fc91467497e6e4eb64336b/draft/DRAFT_TEMPLATE#L22-L33 ?

### DIFF
--- a/content/2025-08-06-this-week-in-rust.md
+++ b/content/2025-08-06-this-week-in-rust.md
@@ -45,12 +45,12 @@ and just ask the editors to select the category.
 * [Meilisearch 1.16 — multi-modal embeddings and direct migration API](https://www.meilisearch.com/blog/meilisearch-1-16)
 
 ### Observations/Thoughts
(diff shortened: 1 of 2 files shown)


nellshamrell:
Thank you so much! It would be wonderful if you update the guidance in that file as well!

homersimpsons:
> Thank you so much! It would be wonderful if you update the guidance in that file as well!

I added this guidance, I mentioned `[video]` and `[podcast]` (used in https://this-week-in-rust.org/blog/2015/10/26/this-week-in-rust-102/ for example), I believe we could imagine other kind of contents, for example `[game]`, I'm not sure about this, please feel free to recommend a different wording.

nellshamrell:
This is great! Thank you so much for contributing this!

nellshamrell (approved):

PR #6885, opened 2025-08-10
Observations/Thoughts: Are we Teaching Rust Effectively?

parasyte:
I wanted to make a broad suggestion for improving the quality of learning materials especially for addressing the perception of Rust's learning curve. It is unclear what would make quantifiable improvements, but in this article I explore the possibility that focusing on ownership may not be the most effective way to help students become proficient.

A blog post seemed like the best first step, since this is informal opinion, at best.

### DIFF
--- a/draft/2025-08-13-this-week-in-rust.md
+++ b/draft/2025-08-13-this-week-in-rust.md
@@ -42,6 +42,8 @@ and just ask the editors to select the category.
 
 ### Observations/Thoughts
 
+* [Are we Teaching Rust Effectively?](https://blog.kodewerx.org/2025/08/are-we-teaching-rust-effectively.html)
+
 ### Rust Walkthroughs
 
 ### Research


nellshamrell:
Hello there!

Thanks for submitting this. Most of this article is fine. However, I'm concerned the section "Common Missteps in Rust Learning Material" crosses the line into a rant about specific work. This is fixable, however.

In [our guidelines on what we are looking for and not looking for](https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#what-do-we-look-for-when-considering-whether-to-include-something-in-this-week-in-rust), we say we are not looking for:

"Rants or anything degrading to any part or member of the Community. Rather than submitting an article about what is wrong with something, we would much rather you write something that explains how you'd make it better."

Would you be willing to work on the language in the "Common Missteps in Rust Learning Material" section to make it a little less ranty? The rest of this article is good and, should changes be made to that specific section, I would be happy to include this in TWiR.

parasyte:
@nellshamrell Thank you for the feedback. I took it to heart and made some changes to that section, including giving it a more objective title. Hopefully it reads as less ranty. That was not my intention, but I see what you mean.

nellshamrell:
That is MUCH better! Thank you for being open to feedback! Merging :)

PR #6906, opened 2025-08-15
Added Lessons learned from rewriting the UltraGraph crate 

marvin-hansen:
Added Lessons learned from rewriting the UltraGraph crate to Observations/Thoughts section.

Simple reflection I want to share. LMK if that is the correct section.

Thank you


bdillo (approved):
thanks!

PR #6926, opened 2025-08-21
Fix events section formatting 611

mariannegoldin:
Fixing some formatting retroactively


bnchi:
Thank you @mariannegoldin looks good!

I was discussing with Ben on how can we eliminate these subtle mistakes, and was wondering if having different scripts for linting/formatting/collecting-events make sense, IMO we need to have all of these in _one_ CLI program, and we can eventually maintain that single program or add any other tools that helps us in the process.  

I would love to work on an issue that helps in moving the Python script to Rust and merge it with the rest of our _beta_ tooling so that we can make this a _bit_ more usable, and maybe create an issue to track the progress and help with any work that's still needed on the linter/merger scripts. 

What do you think @bdillo

bdillo:
> Thank you @mariannegoldin looks good!
> 
> I was discussing with Ben on how can we eliminate these subtle mistakes, and was wondering if having different scripts for linting/formatting/collecting-events make sense, IMO we need to have all of these in _one_ CLI program, and we can eventually maintain that single program or add any other tools that helps us in the process.
> 
> I would love to work on an issue that helps in moving the Python script to Rust and merge it with the rest of our _beta_ tooling so that we can make this a _bit_ more usable, and maybe create an issue to track the progress and help with any work that's still needed on the linter/merger scripts.
> 
> What do you think @bdillo

I agree, it would be much nicer in the long term to just have everything in rust in general. I should be able to look at finishing the linter/new event merger thing soon and getting that in a working state, then we can see about moving the meetup api stuff over there as well.

bdillo (approved):

PR #6929, opened 2025-08-21
added sledoview project to tools section

sgchris:
Added a CLI tool called Sledoview. A tool for viewing and editing Sled DB files in the command line. The tool has features like CRUD operations in keys and values, list and search by pattern/regex, convenient autocomplete capabilities and nice colorful output design. 
Project website: https://sledoview.greq.me/
Repo: https://github.com/sgchris/sledoview

### DIFF
--- a/draft/2025-08-20-this-week-in-rust.md
+++ b/draft/2025-08-20-this-week-in-rust.md
@@ -45,6 +45,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


bdillo:
thanks for the contribution, however in the project/tooling updates section we don't include plain links to github repos, see: https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#projectstooling-updates. If you have a write up with some detailed content we would be happy to include that though!

PR #6932, opened 2025-08-21
Adding interview with Nell about Rust at Microsoft

drewcapener:
This ones feels pretty inceptiony... This is my interview with Nell, which I realize that Nell may end up approving lol


bdillo (approved):
couldn't have Nell approving this, conflict of interest! thanks!

PR #6934, opened 2025-08-25
docs: update 2025-08-27-this-week-in-rust.md

beryl678:
add rust-related thoughts and CFP projects.


bdillo:
thank you for the contribution, but I don't think this event is a good fit for TWIR. The CFP Projects section is about asking for contributions to open source projects and this doesn't seem to be that, nor is it rust specific. Would be happy to include any greptimedb issues if you were looking for contributors though!

PR #6935, opened 2025-08-25
Update 2025-08-27-this-week-in-rust.md

beryl678:


bdillo (approved):
I don't think the first link is a good fit for TWIR as it doesn't really have any rust content. For the second link I moved it to the Misc section where I think it fits a bit better. thanks!

PR #6953, opened 2025-08-29
RustCurious 1: Why Rust is Safe

benw:
Hi, this video is an introduction to borrowed references and ownership that I hope you'll find sufficiently novel. It avoids using any Rust syntax, instead walking through the invention of a thought-experiment language based on C. It's based on Aria Desire's way of explaining ownership. The basic structure is:

- Pointers are evil, remove all pointers from C
- `File` example: It's still possible to cause use-after-close on a file :(
- Invent move-only types :)
- Now the `write` function consumes the file :(
- Invent borrowing for function parameters :)
- `Array` example - `push` works great, but no way to get a pointer to an element in-place :(
- Invent regions so we can return a borrowed reference from the `get` function :)
- Can't borrow references to two array elements :(
- Invent shared references :)
- Mutable xor shared - accidentally thread-safe 🎉

### DIFF
--- a/draft/2025-09-03-this-week-in-rust.md
+++ b/draft/2025-09-03-this-week-in-rust.md
@@ -49,6 +49,8 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
+* [video] [RustCurious 1: Why Rust is Safe](https://www.youtube.com/watch?v=lVWiHIVXG2c)
+
 ### Research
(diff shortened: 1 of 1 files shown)


benw:
I'm wondering if the link text should instead be "A novel introduction to ownership and borrowing" or something like that, very happy to be guided by you. Thanks!

PR #6962, opened 2025-08-31
Improve the Meetup script venue detection

bnchi:
Some Meetup events don't have a specified venue. These aren't always virtual or hybrid — organizers may put venue details only in the event description, so it's the editor's job to read the description and decide which section the event belongs in.

Previously the script treated missing-venue events as errors; editors had to inspect error logs, review the event, and write the markdown manually. This PR reduces that work by treating events with no venue as normal events and favoring the group's location over the event location when determining the grouping key. Generated draft Markdown now includes a new "No Venue" section for those events so editors can find and handle them more easily.

### DIFF
--- a/tools/events/meetup-automation/event.py
+++ b/tools/events/meetup-automation/event.py
@@ -76,6 +76,7 @@ class Event:
   virtual: bool
   organizer_name: str
   organizer_url: str
+  no_venue: bool
 
   def __post_init__(self):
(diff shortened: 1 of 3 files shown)


nellshamrell:
@bnchi @mariannegoldin is this good to go?

bnchi:
we need to merge #7038 and #6641 before deciding on this

PR #6967, opened 2025-09-01
Add article: Be Careful Zero-Copying Strings with `serde`

BD103:
Thank you all for working on this, I really enjoy reading this newsletter every week :)


mariannegoldin (approved):
Thanks, we appreciate your contribution, makes our work fun :) Nice serde tip. Approved.

PR #6991, opened 2025-09-08
issue 615: fix bad merge causing incorrect categories

ericseppanen:
Links added in fefce65e25fa got caught in a bad git merge due to removed category headings. Fix this by moving them back where they belonged.


mariannegoldin (approved):
Thank you v much

PR #7015, opened 2025-09-14
Add "The pirate-based logic of Rust shared references"

ais523:
A submission for the "Observations/Thoughts" section of This Week In Rust.

### DIFF
--- a/draft/2025-09-17-this-week-in-rust.md
+++ b/draft/2025-09-17-this-week-in-rust.md
@@ -47,6 +47,8 @@ and just ask the editors to select the category.
 
 ### Observations/Thoughts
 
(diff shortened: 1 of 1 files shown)


ada4a:
Sorry for the spam, but I wasn't able to find literally any other way to contact the author, so I'm doing it here^^

Thank you so much for this article @ais523! It, and the previous one about scoped generics, are easily two of the most interesting articles I've read on Rust:)

ais523:
@ada4a: Thanks for the kind words! (I should probably see if I can find a better way to contact me – this is the second time someone has tried to contact me using a This Week In Rust pull request.)

PR #7019, opened 2025-09-15
Sync perf-triage with latest rustc-perf version

panstromek:
I got some suggestions in the review in rustc-perf repo, so the final text and full document link has changed.


mariannegoldin (approved):
Thanks for the update!

PR #7028, opened 2025-09-16
add Rust Iterator overview

Aloso:
This is a learning resource, so I wasn't sure if _Miscellaneous_ is the best category for this 🤷🏻 

https://iterators.aloso.foo/

### DIFF
--- a/draft/2025-09-17-this-week-in-rust.md
+++ b/draft/2025-09-17-this-week-in-rust.md
@@ -63,6 +63,7 @@ and just ask the editors to select the category.
 
 ### Miscellaneous
 * [August 2025 Rust Jobs Report](https://filtra.io/rust/jobs-report/aug-25)
+* [Rust Iterator overview](https://iterators.aloso.foo/)
 
 ## Crate of the Week
 


mariannegoldin:
Hello, thank you very much for the submission. 
This page, while informative, appears to duplicate some of of the official [documentation](https://doc.rust-lang.org/std/iter/trait.Iterator.html) in the Rust std library docs on `Iterator` and it is not clear on how interpret the implementation categories it provides. A blog or writeup introducing the tool, how it was created and how it should be used would be helpful. Please feel free to submit an update for issue 618; however, we cannot accept the submission for issue 617.

Aloso:
Here's why I find this confusing:

> This page, while informative, appears to duplicate some of of the official documentation

The page is a summary of the Iterator documentation, so it goes without saying that parts of the documentation are duplicated. That's not a bad thing per se. Also, every item links to the official documentation.

> it is not clear on how interpret the implementation categories it provides

Every category has a short explanation, and the meaning of "Adapters" and "Consumers" is explained in the 2nd paragraph.

> A blog or writeup introducing the tool, how it was created and how it should be used would be helpful.

I thought it would be self-explanatory how to use it. I shared it on Reddit, and nobody had trouble understanding it. I didn't want to write an essay about it, and I didn't think that was required? There have been "cheat sheets" shared on TWIR many times.

Aloso:
cc @mariannegoldin @nellshamrell I would appreciate your input as well.

nellshamrell:
Hi @Aloso - I appreciate you submitting this, but there is not enough original content in it for us to include it in This Week in Rust. If you would like to expand on it, that would be most welcome. One thing that would help would be to create some original code samples that go with each function.

Aloso:
That's an explanation I can live with. Thanks

PR #7045, opened 2025-09-21
remove Dockerfile platform and update to latest python

tzilist:
The platform flag is likely unnecessary anymore as most docker images now publish multiple platforms to use. When running this on a relatively new mac (i.e. an M1), this will cause the code to go through docker hypervisor and the rosetta virtualization platform. This should in theory make things a little faster if you are working on a mac.

Also, [python 3.8 is now EOL](https://devguide.python.org/versions/), so this updates the version to the latest stable release.

### DIFF
--- a/publishing/Dockerfile
+++ b/publishing/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 python:3.8.16-slim
+FROM python:3.13.7-slim
 
 # Must run from base TWIR directory... makefile takes care of this.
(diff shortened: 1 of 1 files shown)


nellshamrell:
Ty! I will take a look at this this weekend to do some testing :)

andreacfromtheapp:
I have prposed a fix for all the dockerfile/makefile issues here: https://github.com/rust-lang/this-week-in-rust/pull/7072

PR #618, opened 2018-03-26
number 227 - added blog post

milancio42:
Hi There! I added a new blog post "Designing a fast CLI join tool with rust".
Thanks,
milancio42

PR #7065, opened 2025-09-24
add community links for #618

opeolluwa:


ericseppanen:
Can we add the rust 1.90.0 announcement? https://blog.rust-lang.org/2025/09/18/Rust-1.90.0/

ericseppanen:
Also, there are a couple of posts this week on the Inside Rust blog: https://blog.rust-lang.org/inside-rust/index.html

We should always include everything that gets posted to the official rust blog(s).

ericseppanen:
I added a commit with the desired changes to this PR.

PR #7070, opened 2025-09-25
fix: issue 6879 - to allow Firefox offline translation tool

andreacfromtheapp:
this was an easy fix: adding <html lang="en"> to https://github.com/rust-lang/this-week-in-rust/blob/master/themes/rusted/templates/base.html -- all pages should inherit this afterwards.

I have run the Makefile and Docker environment to see the tag correctly applied. Someone using Firefox in another lang: could you please confirm thir has the intended behavior?

### DIFF
--- a/themes/rusted/templates/base.html
+++ b/themes/rusted/templates/base.html
@@ -1,5 +1,5 @@
-<!DOCTYPE html>
-<html>
+<!doctype html>
+<html lang="en">
   <head>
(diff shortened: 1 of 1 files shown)


nellshamrell (approved):
Ty so much!

andreacfromtheapp:
My pleasure! :) thanks for TWiR!

PR #7080, opened 2025-09-29
Add blogr v0.3.0 and feedr v0.2.0 announcements

bahdotsh:


bahdotsh:
I'll definitely keep that in mind for next time! Thank you for approving!

PR #7081, opened 2025-09-29
Article: A Rust beginner's thoughts

rosscanning:


ericseppanen:
Thanks for the link!

Unfortunately, I think the article is too basic for our audience. Good luck with your Rust journey!

PR #7095, opened 2025-10-01
Add utsuru: "Go Live" on Discord using OBS, FFmpeg, or anything that supports WHIP.

VincentVerdynanta:
Hi TWiR editors, this is a follow-up to #7064.

I wanted to ask a quick question regarding the project write-up. The project guidelines mention including an introduction, examples, or insights from development. I actually wrote a more detailed introduction to utsuru when I released v0.1.0, and shared that on [Reddit](https://www.reddit.com/r/rust/comments/1lekud7/utsuru_go_live_simultaneously_on_multiple_discord/) and [HN](https://news.ycombinator.com/item?id=44311016) at the time. It didn't make it into TWiR back then, but I'm now submitting with the release of v0.2.0.

Should I link to that original Reddit or HN post instead as the write-up for this submission, or I can keep the GitHub repo link? I'd prefer not to reintroduce the project from scratch in the v0.2.0 changelog, since it might feel redundant or out of place there.

Thanks again for the time, effort, and all the work that goes into TWiR!

PR #7100, opened 2025-10-02
Add utsuru: "Go Live" on Discord using OBS, FFmpeg, or anything that supports WHIP.

VincentVerdynanta:


bdillo:
ty for the contribution, however we don't include just plain repo links in twir per our [policy](https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#projectstooling-updates). I see in #7095 you updated the release page, so I'm just going to swap that out instead of the plain github repo link in this pr.

PR #7114, opened 2025-10-07
Add a Rust course taking place in Gent

wvhulle:
I am organising an in-depth Rust course between November and December 2025.


bdillo:
thank you for the contribution, however we do not include paid workshop-type events in the events section, per the [guidelines](https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#events).

PR #7158, opened 2025-10-19
Add project update: "fzf-make v0.65.0"

kyu08:
Thank you for maintaining TWIR always!

I released v0.65.0 of fzf-make which is a command line tool that executes commands using fuzzy finder with preview window. 
Currently, [make](https://www.gnu.org/software/make/), [pnpm](https://github.com/pnpm/pnpm), [yarn](https://github.com/yarnpkg/berry), [just](https://github.com/casey/just) and [task](https://github.com/go-task/task) are being supported.

I would be happy if you could consider merging this :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ Use the included `new_contribs.sh` script:
 ## Building
 
 To ensure consistency across development setups, we use a [Docker](https://www.docker.com) container-based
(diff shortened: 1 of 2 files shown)


nellshamrell (approved):
Ty!

kyu08:
Thanks for reviewing and the fix!

PR #7169, opened 2025-10-21
Rust LWN articles

ojeda:

Thanks!


ojeda:
One more added, just published.

nellshamrell (approved):
Ty so much!

ojeda:
> Hi @ojeda
> 
> These are very impressive.
> 
> Would you mind moving tem under `Observations/Thoughts` where they are more suited.

You're welcome! Sometimes they have been placed in one and sometimes in the other -- normally I just put them in the same one since these articles are typically about people talking about their work, but what is the difference between the two categories? Thanks!

PR #7188, opened 2025-10-25
add Rust for JavaScript Engineers - Connect-4 Interactivity

tauseefk:
### Add to Rust Walkthroughs - Rust for JavaScript Engineers - Connect-4 Interactivity


mariannegoldin (approved):
TY for the submission - approving. As a heads up, some of the examples should be re-checked to ensure they compile (missing import of Player, missing () on the getter, and index type mixing)

PR #7223, opened 2025-11-04
Remove mistaken stable ICE file dumps FCP

teor2345:
That ticket has been closed as not planned since 2024, not sure how it ended up in the list:
https://github.com/rust-lang/compiler-team/issues/809


nellshamrell:
Ty! Merging this and the update will go out when we publish tomorrow.

PR #7244, opened 2025-11-06
Fix QOTD link in 2025-11-05

vincentdephily:
Missing link recovered from URLO.


mariannegoldin (approved):
TY very much for catching this!

PR #7247, opened 2025-11-07
Update rust meetup lists for automation

mariannegoldin:
Doing some housekeeping; moving some confirmed Rust meetups to the authoritative list, adding some new groups that have been appearing in past issues.

Note: this PR does not account for any groups that are no longer active or deleted. That is good future work.


bnchi:
Thank you!

BTW a quick grep on the events script logs to filter out all the non-valid Rust meetup group should do it to know which ones are no longer active

tools/events/maybe-rust-meetups.json
```diff
@@ -4,25 +4,25 @@
...
     "https://www.meetup.com/bitdevsla/events/",
     "https://www.meetup.com/blockchaincenter/events/",
-    "https://www.meetup.com/boulder-elixir/events/",
-    "https://www.meetup.com/code-mavens/events/",
+    "https://www.meetup.com/blockchaincentersf/events/",
     "https://www.meetup.com/coderrange-endless-programming-languages/events/",
     "https://www.meetup.com/conf42/events/",
     "https://www.meetup.com/data-ai-online/events/",
     "https://www.meetup.com/func-prog-sweden/events/",
-    "https://www.meetup.com/hackerdojo/events/",
```
bdillo: I see a few groups like this and boulder-elixir that just occasionally have rust events being moved to the `rust-meetups` file. The intent of splitting these up was that `rust-meetups` had groups that did rust specific events, so everything would get included by default, whereas in `maybe-rust-meetups` you would have to look through to see what events were rust related.

imo we should either keep the groups that aren't rust specific separate, or just toss everything into a single file and search everything for rust events.

PR #7250, opened 2025-11-09
Update 2025-11-12-this-week-in-rust.md

kirkjkrauss:
Under Rust Walkthroughs, link to article discussing Rust implementations of an algorithm for matching wildcards.  Includes code for ASCII and UTF-8 content, thorough testcase code, and a discussion contrasting the algorithmic approach with the approach taken in prior methods of matching wildcards in Rust.

### DIFF
--- a/draft/2025-11-12-this-week-in-rust.md
+++ b/draft/2025-11-12-this-week-in-rust.md
@@ -48,6 +48,7 @@ and just ask the editors to select the category.
 ### Observations/Thoughts
 
 ### Rust Walkthroughs
+[Matching Wildcards in Rust](https://developforperformance.com/MatchingWildcardsInRust.html)
 
 ### Research
 


nellshamrell:
Hello there! Thanks for submitting this to This Week in Rust! We have lots of regular contributors, so I'm this notice to each pull request we receive this week.

NO ACTION IS NEEDED ON THIS pull request - this is only a heads up for next week and beyond.

On November 13, This Week in Rust will be following the example of the Rust Project and renaming its default branch from "master" to "main". Starting next week, pull requests will need to be to the "main" branch, rather than the "master" branch. For guidance on how to rename your fork's default branch and update your local checkout, please see [this Rust Project blog post for guidance](https://blog.rust-lang.org/inside-rust/2025/10/16/renaming-the-default-branch-of-rust-lang-rust/).

Thank you again so much for contributing to This Week in Rust!

ericseppanen:
I think the code in this article could be improved with a better understanding of Rust. The code uses manual indexing and length checks that could be replaced with slices and iterators, and includes tests that don't use Rust's built-in unit test facility.

As a result, I don't think this article is suitable for TWiR at this time.

PR #7269, opened 2025-11-13
Fix markdown rendering

phil-opp:
The "Compiler Team (MCPs only)" subsection in the current newsletter issue is [not rendered correctly](https://this-week-in-rust.org/blog/2025/11/12/this-week-in-rust-625/#tracking-issues-prs):

image

(Rust subsection included for reference)

This PR makes it an actual heading so that it should be rendered like the "Rust" subsection.

Alternatively, we could add a newline between the title and the list items. I think this would also fix the rendering of the list. Just let me know what you prefer.


U007D:
Ah, this was my mistake.  Good catch and thanks, @phil-opp, for the fix!

LGTM

nellshamrell (approved):
Good catch! Ty!

PR #7276, opened 2025-11-17
Add CFP details for Rustikon 2026 event

wkargul:


wkargul:
@softberries FYI I'm sharing CfP within TWIR

PR #7301, opened 2025-11-20
Fix link to `cargo cat`

AMDmi3:


nellshamrell:
Thanks so much for submitting this fix! Looks like we got it in another pull request as well and merged it in. Thank you anyway, though, and thank you for reading This Week in Rust!

PR #7304, opened 2025-11-20
Add loglens-core and blog post

Caelrith:
Added github repo and blog post for a new, modern rust based log analysis tool.


bdillo:
hey @Caelrith, ty for the contribution, but as @opeolluwa mentioned we don't include plain repo/crate links due to the volume of things we would include, see [this section of the readme](https://github.com/rust-lang/this-week-in-rust/blob/main/README.md#projectstooling-updates). 

for the other link, I don't see any specific rust content in it, so I don't think it's a good fit for TWIR. if you were to write a blog about your tool with more rust content we would be happy to include it!

PR #7305, opened 2025-11-21
Fixed CFP item causing RSS failure

chris-t-jansen:
Fixed a malformed listing for the `Rustikon 2026` item in the `CFP - Events` section of the 2025-11-19 newsletter. This should allow the RSS feeds to build correctly now.


chris-t-jansen:
It seems like there should be some kind of automated test to catch this sort of issue if it's capable of slipping under the radar and causing a failure of the RSS feeds. I don't know enough to suggest where that kind of automatic test should go, but if someone can point me in the right direction, I'd be more than happy to take a crack at it.

nellshamrell:
Good catch! Ty! Merging this in and deploying it.

PR #7340, opened 2025-12-02
Add Clippy Changelog Cat Contest 1.92

blyxyas:
Add Clippy's Changelog Cat Contest for 1.92, see at https://github.com/rust-lang/rust-clippy/pull/16158


nellshamrell:
This is very cute, but I don’t think there’s enough Rust content for this to be in This Week in Rust. Adorable cats, though!

blyxyas:
This might be one of the saddest moment in anime...

PR #7359, opened 2025-12-06
chore: Should we get rid of clippy::manual_try_fold? from blog.veeso.dev

veeso:
Added this article <https://blog.veeso.dev/blog/en/should-we-get-rid-of-clippy-manual-try-fold/> from my blog.

Thank you!


nellshamrell (approved):
Ty!

PR #7368, opened 2025-12-09
Blog post on avx-512 and ... miri

erikjee:
Blog post by https://github.com/folkertdev:

> How do you test code that a CI machine won't run? To support avx-512 SIMD instructions in zlib-rs, we taught Miri how to emulate them.


nellshamrell:
Looks like we got this on3e in another pull request, but thank you!

Issue #7324, opened 2025-11-26
Podman support as a Docker alternative


chris-t-jansen:
I'm not exactly sure where this discussion is most appropriate since it's about TWiR development rather than content, but I wanted to suggest including [Podman](https://podman.io/) support as an optional alternative to Docker, since Podman is totally open-source and an almost 1-to-1 drop-in replacement for Docker. I've personally gotten this to work on two of my own machines (macOS and Windows through WSL), and it was quite literally almost as simple as a "Find and Replace All" command on the `justfile` to swap out `podman` in place of `docker`.

I definitely wouldn't want to disrupt any existing workflows with the added support, so I'd aim for a solution where the Podman support is opt-in, keeping Docker as the default (e.g. `just website` would use Docker, while `just website podman` would execute the same recipes, but using `podman` in place of `docker`). Since `just` supports recipe parameters with default values ([see here](https://just.systems/man/en/recipe-parameters.html)), as well as conditional expressions inside recipes ([here](https://just.systems/man/en/conditional-expressions.html)), I think this should be relatively simple.

I'm happy to write and submit the PR myself, but I wanted to post it as an issue first to make space for discussion. The obvious downside is that it adds some complexity to the `justfile`, which maintainers may not see as being worth it to support what will likely be a relatively small number of contributors. Obviously I'm in favor of the support, and personally I'll continue using Podman either way, but I totally understand if maintainers aren't interested in adding it to the repo in general.

nellshamrell:
Hi Chris! As long as podman is opt in and Docker is still the default, I'm definitely open to this! And this is the correct place for this discussion :)

chris-t-jansen:
Created a PR for this: #7374 

It involved more refactoring than I originally expected, namely on the email side (the `just email` chain of recipes), so I tried to include a detailed write-up of the changes I made in case any of them need altered or are deal-breakers.

PR #7374, opened 2025-12-09
Add alternate container engine support to Justfile

chris-t-jansen:
## Overview

PR for #7324. Adds alternate container engine support to the `justfile` for TWiR. This PR was written with [Podman](https://podman.io/) in mind and was tested with that, but should work with any drop-in Docker replacement. Use it with any recipe that previously called Docker like so:
```
$ just <recipe> <container engine>
```
For example:
```
$ just website podman

$ just email podman
```

> [!WARNING]  
> Though I feel that what I've written is sound, additionally testing is immensely appreciated just to be sure, especially by the TWiR editors since they are the people who will use it the most.

### Docker by default

Docker remains the default container engine, so passing in no arguments (i.e. `just website`, `just email`) will use Docker as it always has. This is achieved by having Docker set as a global Just variable here:
```just
default-container-engine := "docker"
```
Which then gets used as the default value for the newly added `container-engine` parameter on any recipe that used to call `docker`. For example:
```just
# Build container engine image (Docker by default)
container-build container-engine=default-container-engine:
	cd .. && {{container-engine}} build -t twir -f publishing/Dockerfile . && cd -
```

### Cons

A refactor like this usually comes with a downside or two, and this one is no exception. The obvious one here is that the `justfile` is fairly more verbose, with the inclusion of parameters and default values for most recipes. This makes it less easily readable, which may hurt maintainability in the long run.

The other downside has to do with the `optimize-email` recipe. This recipe originally just called a shell script (`create_optimized_email.sh`), which itself explicitly called `docker`. To make it engine-agnostic, I inlined the script into the `justfile` as a [script recipe](https://just.systems/man/en/script-recipes.html). Originally I implemented this as a [shebang recipe](https://just.systems/man/en/shebang-recipes.html), but due to `noexec` issues on WSL (see [here](https://github.com/casey/just/issues/1611) and [here](https://github.com/casey/just/issues/2702)), I had to opt for the unstable `[script(COMMAND)]` attribute instead. The default `COMMAND` of `sh -eu` didn't work for me in testing on WSL, so I opted for `bash` instead. This could cause issues with portability on machines that don't include or symlink a `bash` executable, but I had no trouble when testing it on my Windows machine with WSL and with my macOS machine.

The `optimize-email` recipe issue could also be resolved by refactoring the original script to support command line arguments (e.g. `create_optimized_email.sh -c podman`), but keeping things in the `justfile` seemed a more elegant solution to me.

### Concerns

If you have any questions, comments, concerns, etc., feel free to let me know. As I said in the original issue I opened, I think this kind of compatibility support is important, but totally understand if the TWiR editors aren't interested in it.

### DIFF
--- a/content/2026-01-07-this-week-in-rust.md
+++ b/content/2026-01-07-this-week-in-rust.md
@@ -121,7 +121,7 @@ Are you a new or experienced speaker looking for a place to share something cool
 <!-- or if none - *No Calls for papers or presentations were submitted this week.* -->
 
 * [**RustWeek 2026**](https://sessionize.com/rustweek-2026/) | CFP closes 2026-01-18 | Utrecht, The Netherlands | 2026-05-19 - 2026-05-20
-* [**RustConf 2026**](https://sessionize.com/rustconf-2026/) | CFP closes 2026-02-16 | Montreal, Quebec, Canada | 2026-09-08 - 2026-09-10
+* [**RustConf 2026**](https://sessionize.com/rustconf-2026/) | CFP closes 2026-02-16 | Montreal, Quebec, Canada | 2026-09-08 - 2026-09-11
 
 If you are an event organizer hoping to expand the reach of your event, please submit a link to the website through a [PR to TWiR](https://github.com/rust-lang/this-week-in-rust) or by reaching out on [Bluesky](https://bsky.app/profile/thisweekinrust.bsky.social) or [Mastodon](https://mastodon.social/@thisweekinrust)!
 
--- a/content/2026-01-14-this-week-in-rust.md
+++ b/content/2026-01-14-this-week-in-rust.md
@@ -1,0 +1,459 @@
+Title: This Week in Rust 634
+Number: 634
+Date: 2026-01-14
+Category: This Week in Rust
+
+Hello and welcome to another issue of *This Week in Rust*!
+[Rust](https://www.rust-lang.org/) is a programming language empowering everyone to build reliable and efficient software.
+This is a weekly summary of its progress and community.
+Want something mentioned? Tag us at
+[@thisweekinrust.bsky.social](https://bsky.app/profile/thisweekinrust.bsky.social) on Bluesky or
+[@ThisWeekinRust](https://mastodon.social/@thisweekinrust) on mastodon.social, or
+[send us a pull request](https://github.com/rust-lang/this-week-in-rust).
+Want to get involved? [We love contributions](https://github.com/rust-lang/rust/blob/main/CONTRIBUTING.md).
+
+*This Week in Rust* is openly developed [on GitHub](https://github.com/rust-lang/this-week-in-rust) and archives can be viewed at [this-week-in-rust.org](https://this-week-in-rust.org/).
+If you find any errors in this week's issue, [please submit a PR](https://github.com/rust-lang/this-week-in-rust/pulls).
+
+Want TWIR in your inbox? [Subscribe here](https://this-week-in-rust.us11.list-manage.com/subscribe?u=fd84c1c757e02889a9b08d289&id=0ed8b72485).
+
+## Updates from Rust Community
+
+<!--
+
+Dear community contributors:
+Please read README.md for guidance on submissions.
+Each submitted link should be of the form:
+
+* [Title of the linked Page](https://example.com/my_article)
+
+If you add a link to a non-text content please prefix it with `[video]` or `[audio]`:
+
+* [video] [Title of the linked video](https://example.com/my_video_article)
+* [audio] [Title of the linked audio file](https://example.com/my_podcast)
+
+If you don't know which category to use, feel free to submit a PR anyway
+and just ask the editors to select the category.
+
+-->
+
+### Official
+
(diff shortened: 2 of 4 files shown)


nellshamrell:
Thank you so much! I will review this this weekend!

nellshamrell:
Hi there! Just tested this with Docker and WSL and I'm getting this error when running "just website"

image

Is there a dependency of some sort that I need to install?

chris-t-jansen:
My guess is that you need a newer version of `just`. `just --version` should tell you what version you are using, and I believe you'll need to use version 1.33.0 or newer. I was using 1.43.1 on my macOS machine, and updating for me was as simple as running `cargo install just` again (God bless `cargo`).

`unstable` is a `just`-specific setting that allows the use of experimental features within the script (see [here](https://just.systems/man/en/settings.html)). The unstable feature here is the `[script(COMMAND)]` attribute (see [here](https://just.systems/man/en/script-recipes.html)), which I used because of issues with WSL. You can read the detailed explanation in the "**Cons**" section of the PR, but if it ends up being a blocker for the PR being merged, I can get around it by refactoring the `create_optimized_email.sh` script to take commandline args.

chris-t-jansen:
Looks like the `clean-website` recipe got removed from the `just website` workflow in a recent commit (9101984), but I wasn't sure if that was intentional so I resolved the conflict for this PR by keeping the workflow as it was. If it should get removed to align better with the current `justfile`, I'm happy to change it.

nellshamrell:
Good catch! I'm going to check with that maintainer to see whether that removal was intentional. Also I apologize for taking so long to review this, life got busy, but I will get it done soon.

nellshamrell:
that was indeed an accident, it's back now!

chris-t-jansen:
> that was indeed an accident, it's back now!

Glad I could help!

> I apologize for taking so long to review this, life got busy, but I will get it done soon.

Truly, it's no problem! This is a quality-of-life PR more than anything, it kinda defeats the purpose if it makes your life more difficult. I imagine that when working on a weekly-release publication, it's tough to make time for something that *doesn't* contribute to the next release. Whenever it gets reviewed is great, it's not urgent.

nellshamrell:
This overall looks good to me. @tzilist, since you are using Podman, would you be interested in testing this one out?

chris-t-jansen:
I realized that the "Building" section of the README should also be updated to include instructions about using Docker alternatives, so I added a commit updating that section with instructions using Podman as an example alternative.

As I was updating the "Building" section, I noticed a couple other bits that seemed outdated or slightly misleading, so I ended up updating those as well, though that was perhaps outside the scope of this PR. If it's preferred, I could move that to its own PR.

nellshamrell:
Ty so much! I do believe that is in the scope of this PR - I've designated some time tomorrow to do a review of this :)

tzilist:
@chris-t-jansen most of this seems to be working thankfully! Unfortunately, when i run `just generate-email podman` or `just generate-website podman`, I am running into this bug

```
podman run -it -e USE_EMAIL_THEME=1 -v /Users/tzilist/projects/this-week-in-rust/publishing/output-email-format:/usr/twir/output twir:latest
[03:40:50] CRITICAL FileNotFoundError: [Errno 2] No such file or directory: '/usr/twir/output/atom.xml'                       __init__.py:550
error: Recipe `generate-email` failed on line 94 with exit code 1
```

Would you mind taking a look when you have a sec? Thank you for all of your work on this so far, this is quite nice!

chris-t-jansen:
@tzilist That's interesting! And unfortunate, given that those are kind of the core functions of the `justfile`. Sadly, I'm not seeing that on either of my machines, which makes it tough to troubleshoot. It looks like it's Pelican that's throwing the error, as you can see the line it's referencing [here](https://github.com/getpelican/pelican/blob/bb10d286a6ac2f283d332f7e80c5655b9997f827/pelican%2F__init__.py#L550), but the question is what changed with the `justfile` that caused it. Would you mind trying a couple things to help me narrow it down?

1. Do `just generate-email podman` and `just generate-website podman` throw the same `CRITICAL FileNotFoundError: [Errno 2]` error, and do they throw it every time consistently?

2. Will you try commenting out the feed-generating settings in `pelicanconf.py` and see if it still happens?
```python
DEFAULT_LANG = u'en'

'''
FEED_DOMAIN = SITEURL
FEED_ALL_ATOM = 'atom.xml'
FEED_ALL_RSS = 'rss.xml'
FEED_MAX_ITEMS = 4
CATEGORY_FEED_ATOM = 'categories/{slug}/atom.xml'
CATEGORY_FEED_RSS = 'categories/{slug}/rss.xml'
RSS_FEED_SUMMARY_ONLY = False
'''

DEFAULT_PAGINATION = 10
```

3. Will you separately try modifying the `Dockerfile`'s `CMD` instruction with `--verbose` and sharing the output of the `just` commands? Should look something like this:
```dockerfile
CMD ["pelican", "--delete-output-directory", "--verbose", "content"]
```

Apologies for the laundry-list of requests! Hopefully it sheds some light on what's going on, because after that my next best idea is to try bumping the Pelican version to see if it's something that got fixed later, but that's more of a hopeful solution than a logical one.

PR #7375, opened 2025-12-09
TWIR submission: "Postfix Macros and `let place`"

Nadrieril:
This is a blog post I wrote to explore a feature proposal I've been thinking about.


nellshamrell (approved):
Ty!

PR #7390, opened 2025-12-15
Fix PR title truncated by GitHub

teor2345:
The script works correctly, but GitHub itself has moved the last few words of the title into the first line of the PR description.


nellshamrell (approved):
Ty!

PR #7434, opened 2025-12-30
draft: Add next Rust Girona on-line events

ifraixedes:
Add Rust Girona on-line events for January 2026.


mariannegoldin:
Thank you very much for the submission! It has just been in included in the bulk update in #7442. Closing in lieu of that PR.

Issue #5029, opened 2024-01-04
Include the spoken language in the event listings - at least in the virtual events


szabgab:
I think it would be a good idea to include the language of the presentation in the event listings. At least for the virtual events.

I can guess that the events in the USA and Canada are in English, but what language is used at the event in Nuremberg, Germany.

I recently submitted one that has no physical location it is just  virtual event. I tried to emphasize that it is in English and it was included in the anchor text. It is fine, but maybe this should be somehow formalized in the listings.

bennyvasquez:
This sounds amazing, but would definitely be an increase in work required. @mariannegoldin what do you think?

szabgab:
I'd go further with this suggestion to change the virtual events to include language and time (in UTC and/or in local time) instead of the location which is really not that important for virtual events.

If you think it is a good idea, I can volunteer to handle this.

mariannegoldin:
@szabgab thank you so much for your contributions to TWIR and the community, and for the salient suggestion.  I agree that the location/language addition for each event would be helpful for readers when deciding if they want to attend an event. Adding this info is a balancing act between added effort and sustainability for what would be a manual process on a weekly basis. I am open to suggestions!

cc: @JoelMarcey  for a second opinion


Tl;dr: I think we could use an automated/AI-based solution, as making these updates will require added manual work for the events editors. 


I'll answer in two parts:
* **local time:** this would be useful, but would require querying the meetup API to be sustainable. I am working with some student contributors on an automation to pull this data from the Meetup API programmatically, but it is not yet finished. Otherwise, we would need to update this manually. 
* **language:** this is a harder one because the [Event](https://www.meetup.com/api/schema/#Event) object in the Meetup API does not have a language attached. So, the language indication would need to be made through subjective interpretation, again, manually. 

I could envision an approach for addressing both items above would be to create some kind of AI-based script/scrape to run over the compiled URLs, and provide the location and language for each event. It could be done with a smart agent, BabyAGI, etc. The process might be that I would have a script to run over my draft section, which would recommend the local time and language for each event. I would cross-check each suggestion before adding it. Over time, since I have a working memory of the various groups, the cross-checking would be minimal. 

I welcome suggestions/ideas on how to approach this in the future.

JoelMarcey:
I think the only practical way to have this work is if the events themselves specified the language in some meaningful format that is easily discoverable or parsed.

PR #7475, opened 2026-01-08
update title of virtual event to include time and language

szabgab:
I've updated the titles in Meetup and updating them here too to include the UTC time and the language.

This might be a solution to the issue #5029 if you think it is ok and if other virtual event organizers also adopt it.


nellshamrell (approved):
Ty so much!

PR #7480, opened 2026-01-09
Rust LWN articles

ojeda:

Thanks!

Note: perhaps a code span could be used for both `READ_ONCE()` and `WRITE_ONCE()`, but I didn't know if you want/do that in the titles here and the LWN's title didn't do it anyway (they do it in the body of the article, though), so I left it as-is.


ericseppanen (approved):
Thanks!

PR #4272, opened 2023-04-16
Temporarily remove search functionality

andrewpollack:
Context: #4270 #4271

Search functionality loading is blocking the front page from loading for readers on mobile/tablet devices. This PR temporarily disables search functionality until longer term fix can be implemented.

Issue #4271, opened 2023-04-16
Prevent slow "search"-functionality loading from blocking front page loading


andrewpollack:
Context: #4270

Today, the search functionality takes time to load, preventing readers from using the front page. Immediate fix is to disable the search functionality, but longer term fixes could be:
* Fix described slowness behavior in #4270
* Convert search loading functionality a non-blocking call
* Move search functionality to its own page (thus unblocking front-page from search related slowness)

chris-t-jansen:
Forgot to mention that #7485 would theoretically close this issue, but that PR is a draft at this time.

Issue #4270, opened 2023-04-16
Slow homepage response on iOS / iPadOS Safari


andyeff:
A lengthy delay in the This Week in Rust homepage becoming interactive on iOS has been observed, on both an iPhone 13 and an iPad (6th-gen).

The delay can be up to 10-11 seconds before the page is fully rendered and links become interactable.

An example screen capture showing this can be found in the embedded video at https://qth.social/@andyf/110209113521888114 (or directly at https://qth.social/system/media_attachments/files/110/209/109/931/381/825/original/e0336127f25696fb.mp4 )

From what I've been able to observe, might it be an issue with a `search-index.st` being sent to the client, which looks to be a 15MB blob, and that the processing of this index is what is blocking a main thread in Safari?

Everything works perfectly on desktop safari and firefox, on both Mac and Windows.

andrewpollack:
Looking into this now, thank you for raising this issue! Indeed slowness is coming from loading the search functionality

To ensure the frontpage experience is smooth across different reading devices, we've temporarily deactivated the search feature (reflected now on the website), and created a followup issue at #4271 . I've confirmed on my end that this has resolved the slowness, but please let us know if the slowness continues on those devices!

PR #7485, opened 2026-01-11
Implement site search with Pagefind

chris-t-jansen:
I stumbled across #4271 while looking through old issues and thought it looked interesting, so I decided to take a shot at it.

Rather than a ready-to-merge implementation, this PR is more a proof-of-concept, hence I've opened it as a draft PR. It proves the possibility and outlines the general process, but leaves a fair bit to be desired in terms of implementation. It also begs the question of whether TWiR wants a search function at all since it's been going fine without one for 2.5 years now, but I leave that question to the appropriate authorities.

I tested the updated build process in this PR on macOS and Windows through WSL. I tested the website on Chromium on a MacBook Pro, on Firefox on a Windows PC, on Firefox on Android, and on Safari on a simulated iPhone on a MacBook. In my experience, building the website took an additional 3-5 seconds with the indexing step to enable search functionality, and I didn't notice any additional or unusual slowness when loading the website.

If you've got any questions, comments, concerns, etc., don't hesitate to let me know.

### Problem

The TWiR site previously had search functionality provided by [Pelican's search plugin](https://pypi.org/project/pelican-search/), which under the hood is powered by [Stork](https://stork-search.net/). However, due to an issue causing [website loading issues on iOS/iPad devices](https://github.com/rust-lang/this-week-in-rust/issues/4270), the search functionality got [temporarily removed](https://github.com/rust-lang/this-week-in-rust/pull/4272), and hasn't been revisited since as far as I can tell.

In diving into this issue, I began by reenabling the old search functionality and trying to build the site, but after spending an afternoon playing whack-a-mole with numerous errors, I still couldn't get it to build. I didn't find that too surprising, however, as Pelican's search plugin was [last updated 2 years ago](https://github.com/pelican-plugins/search), and [Stork officially ceased development 3 years ago](https://github.com/jameslittle230/stork/discussions/360). At this point I was left to find an alternative, of which the developer of Stork helpfully provided several possible options. After looking over the recommendations, I landed on [Pagefind](https://pagefind.app/) as the best-looking alternative from my perspective.

### Solution

[Pagefind](https://pagefind.app/) is a static site search binary that, like Stork, is written in Rust (yay!). Because static sites, by definition, don't have a server to handle search queries live, Pagefind works by "indexing" the site at build time and bundling that output into the static site, which a little JavaScript magic on the client's side turns into a functional search bar. Admittedly, it's pretty nifty.

Pagefind conveniently has a [Python wrapper package on PyPi](https://pypi.org/project/pagefind/), which makes installation a breeze. However, Pagefind is a little too new for the `python:3.8.16` image that the build container is derived from, so I had to bump it up to the `python:3.9.25` image. I know that can theoretically cause breaking compatibility issues, though I didn't encounter any in testing, even when bumping it as high as `python:3.11.14`.

The Pagefind settings out of the box are pretty reasonable, but adding a few tags to the site's HTML templates gives it a big boost, such as always sorting results by the most recent issue date.

### Cons

With a significant feature implementation like this, there will always be some negative consequences, no matter how much I enjoy saying "zero-cost abstraction" into the mirror. The biggest *potential* problem is likely the image bump from `python:3.8.16-slim` to `python:3.9.25-slim` due to the possibility of compatibility issues, though as I said I didn't encounter any.

Additionally, the indexing process does add time to every build of the website, around 3-5 seconds in my experience. Some kind of `--dont-index` flag could probably be added to the justfile to suppress the indexing step during development if it's a major issue, though it never bothered me.

Probably the biggest current issue is simply the design. I used the prebuilt search bar bundled with Pagefind and inserted it where the old search bar was before removal. The bar's horizontal spacing is a little awkward, and the search result text doesn't seem to change color in dark mode, making it nigh unreadable in that situation. Furthermore, it might make more sense to put the search function on its own page and link to that page in the header. All worthwhile questions, but I decided that trying to address them before starting a discussion was putting the cart before the horse, so I left the implementation as-is.

### DIFF
--- a/publishing/Dockerfile
+++ b/publishing/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9.25-slim
+FROM python:3.8.16-slim
 
 # Must run from base `twir` directory... justfile takes care of this.
 WORKDIR /usr/twir
@@ -23,6 +23,9 @@ RUN pip3 install -r requirements.txt
 RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
 RUN apt-get install -y nodejs
 RUN npm install -g sass juice
+
+RUN curl -L -o pagefind.tar.gz https://github.com/Pagefind/pagefind/releases/download/v1.4.0/pagefind_extended-v1.4.0-aarch64-unknown-linux-musl.tar.gz
+RUN tar -xvf pagefind.tar.gz
 
 # pelican setup
 COPY content content
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,3 @@ Markdown==3.3.7
 markupsafe==2.0.1
 pelican==4.7.1
 pelican-webassets==2.0.0
-pagefind[extended]==1.4.0


chris-t-jansen:
I did some testing, and there is a way to get around bumping the build container's version, though I can't say I'm a huge fan of it. Basically, instead of installing the Python wrapper package for Pagefind, you can download a precompiled binary directly from [Pagefind's GitHub releases](https://github.com/Pagefind/pagefind/releases). However, the catch here is that, instead of simply downloading a single package and calling it a day, you have to download the correct binary for the architecture of the image, which is dependent on the machine you're building on.

For example, when building on my Apple Silicon macOS machine, the architecture of the container is `arm64`, but when building on my Windows machine through WSL, the architecture is `amd64`. Handling just those two cases requires inserting a somewhat verbose `case` statement into the `Dockerfile`:

```dockerfile
RUN apt-get install -y nodejs
RUN npm install -g sass juice

# determine arch type and install matching Pagefind binary
RUN dpkgArch="$(dpkg --print-architecture)"; \
    case "${dpkgArch##*-}" in \
		amd64) export PAGEFIND_ARCH_DL_LINK="https://github.com/Pagefind/pagefind/releases/download/v1.4.0/pagefind_extended-v1.4.0-x86_64-unknown-linux-musl.tar.gz" ;; \
		arm64) export PAGEFIND_ARCH_DL_LINK="https://github.com/Pagefind/pagefind/releases/download/v1.4.0/pagefind_extended-v1.4.0-aarch64-unknown-linux-musl.tar.gz" ;; \
	esac; \
    curl -L -o pagefind_extended.tar.gz $PAGEFIND_ARCH_DL_LINK; \
    tar -xvf pagefind_extended.tar.gz;

# pelican setup
COPY content content
COPY plugins plugins
```
Once that's properly installed, then you run the Pagefind binary directly from the `justfile` instead of through Python:
```justfile
podman run -it \
	-v {{justfile_directory()}}/output-website:/usr/twir/output \
	twir:latest \
	./pagefind_extended --site /usr/twir/output
```

Personally, I don't like the approach as much as the Python wrapper for the simple reasons that it seems more prone to failure and more difficult to maintain. Upgrading the dependency is also a thoroughly manual process, since the names of the precompiled binaries aren't guaranteed to stay the same between releases. However, if bumping the build container version from `python:3.8.16` to `python:3.9.25` is a non-starter, then there is a way around.

PR #7488, opened 2026-01-11
Add link for "Verify Arch Linux artifacts using VOA/OpenPGP"

dvzrv:
Add a link to an article explaining in-depth the use of the Rust-based `voa` CLI for the verification of Arch Linux artifacts (e.g. package files and installation medium) using the OpenPGP backend and a dedicated voa configuration file.

### DIFF
--- a/draft/2026-01-14-this-week-in-rust.md
+++ b/draft/2026-01-14-this-week-in-rust.md
@@ -44,6 +44,7 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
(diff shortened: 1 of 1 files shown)


ericseppanen (commented):
Thanks for the submission. I don't see any Rust content in this article, so I don't think it's a good fit for TWiR.

dvzrv:
> Thanks for the submission. I don't see any Rust content in this article, so I don't think it's a good fit for TWiR.

Hmm, what do you mean? :thinking: 
The project is written in Rust: https://gitlab.archlinux.org/archlinux/alpm/voa

ericseppanen:
The project may be written in Rust, but the linked article includes no Rust content and isn't targeted at Rust developers. It's a grey area, but we do our best to keep TWiR focused. Also, this project is already mentioned in the other Arch blog post included in this issue.

PR #7490, opened 2026-01-11
Add `diesel-guard v0.4.0` to Project/Tooling Updates

ayarotsky:


ericseppanen (commented):
Thanks for the link!

The TWiR community links section uses a style of one link per line, and discourages links to plain github repositories. I'll add it to the issue as a single link.

PR #7494, opened 2026-01-12
Add `What is maintenance, anyway?` blog post

Kobzol:
https://blog.rust-lang.org/inside-rust/2026/01/12/what-is-maintenance-anyway/


Kobzol:
Ok, got it :)

PR #7521, opened 2026-01-15
draft: Add next Rust Girona on-site event

ifraixedes:
Add the next Rust Girona on-site event and delete the ones which are in the past in regards to the date that this newsletter draft will be sent.


bdillo (approved):
ty!

PR #7542, opened 2026-01-21
Add Tel Aviv Code Mavens event on January 26

szabgab:


bdillo:
just merged #7540 which included this, ty though!

PR #7586, opened 2026-01-30
Add s2-lite to Project/Tooling Updates

infiniteregrets:


nellshamrell:
Hi @infiniteregrets! This sounds like a cool project! We don't usually link directly to Hacker news or Reddit posts. Would you be open to linking to release notes or blog post with similar content?

shikhar:
hi @nellshamrell, would this work: https://github.com/s2-streamstore/s2?tab=readme-ov-file#s2-lite

PR #7599, opened 2026-02-04
add Miri talk

RalfJung:
I hope submitting a PR like this is the right way to do this :)


nellshamrell:
This is indeed the proper way! We did get this in another pull request, so I'm going to close this one, but ty for making sure we saw it!

PR #7622, opened 2026-02-07
Add easy-ssh to project updates

DarkLichCode:
Hello! I'd like to add easy-ssh to this week's project updates. It's a high-level SSH client wrapper for Rust designed with a focus on ergonomics and safety through the Builder pattern. It also includes a fail-fast validator for OpenSSH private key formats to help users avoid common libssh2 pitfalls.

### DIFF
--- a/draft/2026-02-11-this-week-in-rust.md
+++ b/draft/2026-02-11-this-week-in-rust.md
@@ -45,6 +45,7 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


ericseppanen:
Thanks for the submission! We don't usually include links that are solely to a GitHub repo. (You can see our [guidelines for this section](https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#projectstooling-updates) for more info).

We like to feature blog posts or other writings about projects, so if you feel like writing up something about your project that might be interesting to other Rust developers, we'd be happy to consider including that!

PR #7626, opened 2026-02-08
Fix list format, state acronyms in 2026-01-28

teor2345:
The list formatting for accepted RFCs isn't rendering correctly on the website, it all ended up as one paragraph.

Australia uses 3-letter state acronyms, two letters is an Americanism. This is fixed in both 2026-01-28 and 2026-02-04, but for some reason I can't change the PR title.

Another alternative would be to delete the Australian states entirely, they're not really used the same way US states are.


mariannegoldin (approved):
Thanks very much for the correction. We appreciate it!

PR #7651, opened 2026-02-13
Add custom 404 page

coolreader18:
When TWIR gets posted to r/rust, there are often comments about it being a dead link, because the reddit post goes up before the blog post. I figure that with a custom 404 page which explains that, some of that confusion could be negated.

### DIFF
--- a/pelicanconf.py
+++ b/pelicanconf.py
@@ -50,6 +50,8 @@ scattered about.
 """
 }
 
+DIRECT_TEMPLATES = ['index', 'tags', 'categories', 'authors', 'archives', '404']
+
 PLUGIN_PATHS = ["plugins"]
 
 # Don't add next/previous buttons search functionality for email.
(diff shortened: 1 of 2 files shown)


ericseppanen:
I wonder if instead we should do a two-step publishing process, where we publish a partial issue first, then create the reddit post, then publish the final issue contents.

nellshamrell:
It's always a chicken and egg situation and this is an interesting solution. I'm going to give some more thought on this - though I am going to go ahead and do the publishing process twice today (which is necessary when we do not create the reddit post first) to avoid this confusion.

nellshamrell:
And ty, @coolreader18, for taking the time to put this together!

nellshamrell:
We've updated our publishing process to publish twice (once without the reddit link so we can create the reddit post, then republishing with the reddit link) because it seems to be the least confusing option for the reader. Thank you so much for your thoughtfulness in opening this as a suggestion, @coolreader18! We really appreciate it!

PR #7675, opened 2026-02-18
Add README section about LLM-written articles

ericseppanen:
Add some guidance about LLM-written articles:
- Explain why we discourage articles that were not written by people.
- Request disclosure on such articles.

It doesn't explicitly say we won't accept such articles, or try to pin down exact conditions for when we might accept or reject. I'm not sure that that's necessary at this time?


nellshamrell:
I don't think it's necessary at this time. This is excellent, ty for putting it together!

PR #7689, opened 2026-02-24
Add diesel-guard v0.7.0 to Project/Tooling Updates

ayarotsky:


ericseppanen:
Thanks for the link. Unfortunately, we've decided not to include it.

1. It doesn't convey much information. A one-line changelog is not sufficiently interesting to share with tens of thousands of subscribers.
2. This project has already been linked 3 times in the last two issues of TWiR.

PR #7736, opened 2026-03-06
RustCurious lesson 4: Structs and Resources

benw:
This video walks through syntax and semantics of tuples and structs, with a focus on the distinction between copyable plain-old-data types versus move-only types that manage resources. It also covers methods with `&self` vs `&mut self` vs `self`, modules and private fields, and `String` vs `&str`. Ends with a practice exercise for the viewer to complete in the playground.

### DIFF
--- a/draft/2026-03-11-this-week-in-rust.md
+++ b/draft/2026-03-11-this-week-in-rust.md
@@ -49,6 +49,8 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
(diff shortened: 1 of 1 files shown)


ericseppanen:
We had a publishing glitch this week and the wrong draft was left in place.

It's fixed now-- would you mind moving this change over to `draft/2026-03-11-this-week-in-rust.md`?

benw:
Ah that makes sense. Done! I took the opportunity to update the link title too.

nellshamrell:
@benw closing for now, but feel free to re-open with the corrected link. Ty again!

benw:
Done - reopened as #7772, thanks :)

PR #7745, opened 2026-03-09
Add zk for rust developer series

adamsmo:
Zero knwledge for rust develoeprs, this is first post out of 8 that explains from first principles how zero knowledger proofs work, first one is about polynomials
I wrote this to have no phd math and from rust dev perspective, so there are rust examples for everything, also you can run all examples in the browser + there is a plyaground at the bottom of the page
I also added 3b1b animations to further explain how things work


nellshamrell (approved):
Ty!

PR #7833, opened 2026-03-25
Add Edge.js project update to Rust newsletter

syrusakbary:
We recently launched Edge.js as a way to run Node.js workloads at native-like speeds but fully sandboxed with Wasm.

Thought it could be useful for ThisWeekInRust readers!


opeolluwa:
Sure @syrusakbary!

ty

PR #7911, opened 2026-04-09
add to rust walkthroughs - add video link for rust condvar tutorial

nazmulidris:


bdillo:
hey, thank you for the contribution @nazmulidris, it looks like this video stream is currently ongoing though, I don't think we'd feel super comfortable adding this to the newsletter until we got a chance to at least skim through the finished video. I did watch for a few mins and it does look like something we'd want to include, so please send us another PR for next week!

nazmulidris:
@bdillo Sorry for the issue with the timing. I will resubmit for next week's issue 👍🏽

PR #7920, opened 2026-04-13
Add Meganeura v0.2, Fileman, and Blade XR Asteroids to the project updates

kvark:
Announcing a bunch of projects I was working on, with help from LLMs.


nellshamrell:
Hi @kvark! Thanks for submitting these. It feels like these release notes and posts were at least in part LLM generated. This is ok, but [we do require that LLM use for articles (including release notes) be disclosed in the article](https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#llm-written-articles). Feel free to add those disclosures and resubmit these links to next week's issue.

Thank you!

kvark:
Ah, I see. Only one of these 3 links is pointing to an LLM generated changelog. I've submitted #7946 instead without Meganeura.

kvark:
Also, it is unfortunate that the PR was sitting here for 3 days without feedback. And now you closed it because of this 1/3 issue, while simultaneously publishing the TWiR edition without any of my submissions. I understand it's not an easy job to maintain TWiR, but I feel that contributions to it aren't as welcome as they could be.

PR #7927, opened 2026-04-14
Add RaTeX to Project/Tooling Updates

erweixin:
Every major cross-platform math renderer today runs LaTeX through a browser or JavaScript engine — a hidden WebView eating 50–150 MB RAM, startup latency before the first formula, no offline guarantee. KaTeX is excellent on the web, but on every other surface — iOS, Android, Flutter, server-side, embedded — you're either hosting a WebView or shelling out to headless Chrome.

RaTeX is the same KaTeX-compatible math engine compiled to a portable Rust core, so the same renderer runs natively everywhere — and produces byte-identical output across every target.

### DIFF
--- a/draft/2026-04-15-this-week-in-rust.md
+++ b/draft/2026-04-15-this-week-in-rust.md
@@ -44,6 +44,7 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hello @erweixin! We unfortunately do not include links that are solely to GitHub repos or crates.io pages (see [our guidelines](https://github.com/rust-lang/this-week-in-rust?tab=readme-ov-file#projectstooling-updates) for more information. Feel free to write this up as a blog post (or do a write up in release notes) and resubmit it to next week's issue!

PR #7933, opened 2026-04-14
Added NDC Techtown under CFP - Events

hpenne:


nellshamrell:
Hi @hpenne! Unfortunately we are publishing this on April 15 and I see that the CFP closed on April 14. However, if you decide to extend it, feel free to submit it again next week!

PR #7949, opened 2026-04-16
Add mtp-mount tooling update

vdavid:
Added my new pure-Rust FUSE mount tool for MTP devices, built on mtp-rs.


tzilist:
Thank you!

PR #7932, opened 2026-04-14
Add eilmeldung to Project/Tooling updates

christo-auer:

PR #7954, opened 2026-04-18
Added write-up on learning rust through an LLM to build a TUI RSS reader

christo-auer:
Added link to my write-up: #7932

Note: I've extended my write-up significantly. I hope it is interesting. 

I am not sure **where** to actually add it: I've added it to *Observations/Thoughts* though it may also fit in *Project/Tools Update* maybe? I rely on your expertise here. Just tell me and I move it.

Thanks!

### DIFF
--- a/draft/2026-04-22-this-week-in-rust.md
+++ b/draft/2026-04-22-this-week-in-rust.md
@@ -47,7 +47,7 @@ and just ask the editors to select the category.
 
 ### Observations/Thoughts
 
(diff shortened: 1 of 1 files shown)


opeolluwa (commented):
Hi @christo-auer, we meet again! 

It's good to read your update. 

I think we can reference `https://github.com/christo-auer/eilmeldung/blob/main/docs/llm-development.md` in place of your current URL, as it better captures the intended semantics.

Ty!

christo-auer:
@opeolluwa I've updated the URL. Thanks!

opeolluwa (approved):
Ty!

PR #7975, opened 2026-04-23
Danube Messaging key_shared subscription

danrusei:
Key-Shared dispatch in Danube Messaging, routing messages by Key with consistent hashing


opeolluwa (approved):
ty!

PR #7983, opened 2026-04-24
Add free book 'Rust Projects - Write a Redis Clone'

lgiordani:
Hi!

This is my first contribution here, I hope I added the link to the correct section.
The book is free both in HTML and PDF forms. Since downloading the PDF version requires an email I added here only the HTML version, that is freely accessible.

Thanks!


nellshamrell (approved):
Ty!

PR #7998, opened 2026-04-28
draft: Remove a Rust Girona online event

ifraixedes:
On that date we have an on site event, which is already in the Europe section, we don't have any virtual, so remove it.


bdillo (approved):
thanks!

PR #7946, opened 2026-04-16
Add Fileman and Blade XR Asteroids to the project updates

kvark:
Related to #7920

PR #8003, opened 2026-04-29
TWIR-649: Add Blade XR Asteroids to project updates

kvark:
Evolving from #7946 and #7920


nellshamrell (approved):
Thank you! I appreciate the revisions!

PR #8014, opened 2026-04-30
Fix incorrect Australian state/territory codes

chris-morgan:
Some countries have two-letter state/territory codes, such as the United States of America and India;
but Australia’s aren’t fixed width.

It so happens that the western ones are all two letters: WA, NT, SA;
but the eastern ones are all three letters: QLD, NSW, ACT, VIC, TAS.

This has been bothering me for a long time.
It took me *way* too long to finally get round to doing this!


nellshamrell (approved):
Ty so much!

PR #8019, opened 2026-05-04
Add kache 0.2.0 to Project/Tooling Updates

jleni:
`kache` is a zero-copy, content-addressed Rust build cache. Drop-in `RUSTC_WRAPPER`; restores cache hits via hardlinks instead of copies. Local-first, blake3 content-addressed, optional S3 sync.

Release notes: https://github.com/kunobi-ninja/kache/releases/tag/v0.2.0


nellshamrell (approved):
Ty!

PR #8025, opened 2026-05-05
Add 'The Sync bound nobody asked for' to Observations/Thoughts

verrchu:
Self-submission of a short Rust post for the 2026-05-06 issue.

The post explains how `&self` on an async trait method whose returned
future must be `Send` implicitly forces `Sync` on the impl type, and
why switching to `&mut self` is usually the cleaner fix than reaching
for a `Mutex`/`RwLock`.


opeolluwa (approved):

nellshamrell (approved):
Ty!

PR #8007, opened 2026-04-29
Add rkik-nts 1.0.0 release note to Rust draft

aguacero7:
Added rkik-nts v1.0.0 release to Project/Tooling Updates.

rkik-nts is a high-level Rust Network Time Security (RFC 8915) client library for authenticated NTP queries.

PR #8027, opened 2026-05-05
Add rkik-nts 1.0.0 release to Project/Tooling Updates

aguacero7:
>This PR is related to this one : https://github.com/rust-lang/this-week-in-rust/pull/8007

Hi,
I adjusted the release notes to match your guidelines,  I hope this is closer to what you had in mind.

`rkik-nts` is a high-level Rust Network Time Security (RFC 8915) client library for authenticated NTP queries, I initially built it for integration into my own project, `rkik`, but designed it as a reusable library for others as well.

One of the major initial motivations behind the crate is that there are not many working NTS client libraries available today. Most of the NTS work I found was either server implementations, daemonized client implementation or more often not exposed as a stable, reusable library interface.

### DIFF
--- a/draft/2026-05-06-this-week-in-rust.md
+++ b/draft/2026-05-06-this-week-in-rust.md
@@ -47,6 +47,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


nellshamrell (approved):
Ty!

PR #8030, opened 2026-05-05
Add "Introduction to database access using Rust SQLx" event

szabgab:


nellshamrell:
Looks like we got this in another pull request, but thank you!

PR #8087, opened 2026-05-14
2026-05-13: Fix QOTW link

vincentdephily:
The original file was removed from the repo. The new link points to the (AFAICT) last commit before deletion.


vincentdephily:
Doc deletion commit is https://github.com/oven-sh/bun/commit/10cfa0b237f8dd9143258800cb25b1c791d5b6b6. Alternatively, the file still exists at https://github.com/oven-sh/bun/blob/claude/r2-phase2-migration/docs/rust-rewrite-plan.md

So many branches and commits, this repo is hard to navigate. The `claude/phase-a-port` branch looks deleted now ?

Just in case, pinging @dylan-conway and @Jarred-Sumner in case they would rather not see a precise link in TWIR.

tzilist:
Thank you!

PR #8096, opened 2026-05-18
draft: Update Rust Girona meetups

ifraixedes:
Drop passed events and add the next scheduled Rust Girona events.


KannanPalani57:
Thank you!

PR #8121, opened 2026-05-25
Add false sharing blog post and snarf linter to issue 653

cong-or:
- Rust Walkthroughs: https://cong-or.xyz/false-sharing-cache-lines.html                                                                                                                     
- Project/Tooling Updates: https://github.com/cong-or/snarf


bdillo (approved):
ty! we don't accept plain github links in tooling/updates though, so I removed that link. I included the walkthrough though.

PR #8132, opened 2026-05-26
add serde-const-default v0.1 to Project/Tooling Updates

ifsheldon:
Hi! this adds the release note of my tool `serde-const-default` which helps remove boilerplate when using const values as field defaults with `serde`. Thanks!


bdillo (approved):
ty!

PR #8133, opened 2026-05-26
Add observation on Physical AI and world models

lxsaah:
Added a new observation about Physical AI needing a typed world model.

Thanks for your work @ TWIR!


bdillo (approved):
ty!

PR #8138, opened 2026-05-27
Add Laissez-Faire Errors link to Rust Walkthroughs

RPG-Alex:
I am not sure if this best for Observations or Walkthroughs as it is a very simple walk through of using errors but is also my opinion on when best to write them. Please feel free to move to Observations if it seems a better fit there.


bdillo (approved):
ty!

PR #8147, opened 2026-05-28
Add webrtc-doctor announcement to Project/Tooling Updates

jamalag:
Adds a project announcement for webrtc-doctor, a single-binary Rust CLI for diagnosing WebRTC connectivity (STUN/TURN/TURNS/signaling/ICE/DTLS). Apache-2.0. The linked article explains the design decision to hand-roll the STUN/TURN wire codecs against the RFCs rather than depend on a high-level WebRTC library, and discusses when that trade-off is correct.


ericseppanen:
Thanks for the submission! Unfortunately, we don't accept links to articles that are only available to account holders.

PR #8153, opened 2026-05-30
Enable warnings for links to github.com and crates.io & suppress existing warnings

### LINKED ISSUE #8148: CI check for commonly-missed guidelines
To start, we want to have CI warn when posting links to github.com or crates.io. 

I have a [draft here](https://github.com/jder/this-week-in-rust/pull/new/more-checks) which does this simple thing. There are currently some recent issues which fail these checks which presumably is annoying (and perhaps will be common since they will probably have false positives) so I think before turning this on we probably want to scope the (new?) checks to only added/edited links, not all links in all recent issues.

As a follow-up, it would be nice for CI to also post a comment back to the PR when it fails (maybe just the first time it fails) to ask the author to take a look.

jder:
Makes the existing CI checks for suspicious links also produce warnings for links to github.com and crates.io, pointing to the [contribution guidelines](https://github.com/rust-lang/this-week-in-rust#projectstooling-updates). 

These are more likely to be false-positives than existing link checks (for example, the recent issue had a couple github.com links). To try and help with this problem, this also introduces a `--since` parameter for the link-checking script which suppresses warnings that already were checked into git. The GitHub Action now passes the merge base (for PRs) or previous HEAD SHA (for pushes) as this parameter. The net result is that contributors should only get *new* warnings, not see old ones. PRs can then be safely merged despite these warnings when appropriate without having them show up over and over. 

In the future we could separate hard warnings from softer ones, and perhaps have the latter just post a comment suggesting folks double check. 

Here is an example PR showing the new warnings: https://github.com/jder/this-week-in-rust/pull/3

cc @nellshamrell 

Fixes https://github.com/rust-lang/this-week-in-rust/issues/8148

I use LLMs regularly while coding, but I understand and vouch for this code.

### DIFF
--- a/tools/inspect_links.py
+++ b/tools/inspect_links.py
@@ -32,6 +32,9 @@ class Warnings:
         if not self.silent:
             self.warnings.append(msg)
 
+    def get(self):
+        return self.warnings
+
     def get_and_clear(self):
         to_return = self.warnings
         self.warnings = []
--- a/tools/inspect_links.py
+++ b/tools/inspect_links.py
@@ -7,12 +7,14 @@ Inspect a set of markdown files, and warn if there are:
 """
 
 import argparse
+from typing import Union
 import bs4
 import logging
 import markdown
 import os
 import re
 import sys
+import pygit2
 import urllib.parse
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


tools/inspect_links.py
```diff
@@ -274,31 +290,43 @@ def main():
                         help="Directory paths to inspect (colon separated)")
     parser.add_argument('--num-recent', default=25, type=int,
                         help="Number of recent files to inspect")
-    parser.add_argument('--num-warn', default=1, type=int,
-                        help="Number of recent files to warn about")
+    parser.add_argument('--since', default=None, type=str,
+                        help="Show only warnings which appear after the given git commit")
```
jder, line 294: One note: I replaced the existing `--num-warn` flag here because I think it was intended to serve the same purpose as the `--since` flag now serves (ie hiding older warnings).

jder:
Also cc @ericseppanen since it looks like you wrote most of this code originally!

nellshamrell (approved):
Thank you so much! This will help a lot!

PR #8158, opened 2026-05-31
RustCurious lesson 8: Generics and Monomorphization

benw:
This video explains generics in detail, covering generic type params and generic const params on functions, types and impl blocks. It ends with a coding exercise to complete in the playground.


ericseppanen (approved):
Thanks!

PR #8159, opened 2026-06-01
Added an article about counterfactuals via the causal monad in Rust 

marvin-hansen:
I added a technical walkthrough about implementing counterfactuals using monadic composition in stable Rust.
This is novel in three ways. First, the project had to implement higher-kinded types in stable rust to enable monadic composition. Second counterfactuals in the conventional sense are only modifying a value, whereas this implementation goes further by enabling the modification of value, state, and context. Third, the bulk of computational causality work is still conducted in Python, and therefore a complete implementation of computational causality in Rust was not available until this project built one.

https://www.deepcausality.com/blog/counterfactuals-via-the-causal-monad/

### DIFF
--- a/draft/2026-06-03-this-week-in-rust.md
+++ b/draft/2026-06-03-this-week-in-rust.md
@@ -51,6 +51,8 @@ and just ask the editors to select the category.
 
 ### Research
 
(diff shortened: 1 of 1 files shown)


ericseppanen (approved):
Thanks!

PR #8160, opened 2026-06-01
Add R1CS vs Plonkish vs AIR walkthrough

adamsmo:
Zero knowledge for Rust developers. This one compares the three ways a ZK system turns a computation into polynomial constraints, R1CS, Plonkish, and AIR. I wrote Fibonacci in all three side by side so you can see exactly what differs between them.

Written from a rust dev perspective with little phd math. The companion repo does each arithmetization in plain Rust, stdlib only, no proving libraries and no dependencies, runs with cargo test, around 1200 lines. There is a narrated audio version of the blog too if you prefere to listen instead of read.

### DIFF
--- a/draft/2026-06-03-this-week-in-rust.md
+++ b/draft/2026-06-03-this-week-in-rust.md
@@ -44,7 +44,6 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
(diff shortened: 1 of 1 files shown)


ericseppanen (dismissed):
The link returns a 404 error for me.

adamsmo:
sorry, now it should work @ericseppanen

adamsmo:
btw, the repo is mentioned in the blogpost [github.com/adamsmo/r1cs-plonkish-air](https://github.com/adamsmo/r1cs-plonkish-air) but I could not add it directly to this PR

nellshamrell (approved):
Ty!

PR #8167, opened 2026-06-02
disable link domain checks

ericseppanen:
This check is too strict (warns on links we currently accept) and is causing CI to fail on the main branch.

CC: @jder


jder:
Thanks! Sorry about that -- will fix with https://github.com/rust-lang/this-week-in-rust/issues/8168 and re-enable then.

PR #8174, opened 2026-06-03
Rust LWN articles

ojeda:

Thanks!

The title of this entry may not seem very Rust relevant, but it does have a bunch of Rust mentions, including a fairly spicy Rust quote... :)


nellshamrell (approved):
Ty!

ojeda:
You're welcome! :)

PR #8177, opened 2026-06-03
Add rustsec.debian.net

kpcyrd:
The project has been around for some time but wasn't properly announced. I have submitted it to r/rust this week:

https://www.reddit.com/r/rust/comments/1tuly3z/debian_rust_security_tracker/

If you want I can write some more how this projects works so it would be suitable for it's own section.

Thanks!


nellshamrell:
Hi @kpcyrd! Thanks for submitting this, I would love to see a write up that goes more into how this project works. Feel free to submit that to next week's issue!

PR #8204, opened 2026-06-10
Add smb2 tooling update

vdavid:
smb2 is a pure-Rust SMB2/3 client library with pipelined I/O, no C dependencies.


KannanPalani57 (approved):
Thank you!

PR #8217, opened 2026-06-14
Add project update: carboxyl 0.1.0-rc

orhun:


opeolluwa (changes requested):
Thanks for submitting this.

Could you format the link with "*" instead of "-"? 
Also the descriptive text for the link is too long; could you limit it to around 5 words?

orhun:
Thanks for the feedback, can you take a look again?

opeolluwa (approved):

PR #8241, opened 2026-06-18
Add cargo-shed 0.1.0 to Project/Tooling Updates

moooossee:
Adds cargo-shed 0.1.0 to Project/Tooling Updates.

cargo-shed is a Cargo subcommand and Rust library for finding dependency bloat, risky feature selections, duplicate crate versions, and safe cleanup opportunities in Rust projects.

### DIFF
--- a/draft/2026-06-24-this-week-in-rust.md
+++ b/draft/2026-06-24-this-week-in-rust.md
@@ -45,6 +45,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
Thank you for submitting this! However, the descriptive text for the link is too long, could you limit it to around 5 words? It also needs to be before the link, right after the title of the link.

We're about to publish, so I'm closing this for now, but feel free to submit this again next week with the requested changes.

PR #8243, opened 2026-06-18
Add ANIMUS to Project/Tooling Updates

ernestoariasdiaz:
ANIMUS is an autonomous Rust system that builds and audits its own knowledge graph, running a local LLM (Gemma 4 E2B) via llama.cpp on consumer GPU. Recent work includes a self-audit that found and corrected metric inflation (1,892 reported nodes → 911 verified) and a four-layer epistemic honesty architecture to prevent confabulated content from entering the graph.

Paper (open access, EN/ES): https://doi.org/10.5281/zenodo.20674981

### DIFF
--- a/draft/2026-06-24-this-week-in-rust.md
+++ b/draft/2026-06-24-this-week-in-rust.md
@@ -45,6 +45,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
+* [ANIMUS v4.0: an autonomous Rust system with a self-auditing knowledge graph, running a local LLM via llama.cpp on consumer GPU](https://github.com/shellhack/animus-ai)
+
 ### Observations/Thoughts
 
 ### Rust Walkthroughs


ernestoariasdiaz:
Thanks so much for including this in the draft! One small request before it goes out: I recently changed my GitHub username from `shellhack` to `ernestoariasdiaz` for professional reasons. The old link (https://github.com/shellhack/animus-ai) still redirects correctly, but if it's not too much trouble, could the link be updated to https://github.com/ernestoariasdiaz/animus-ai before publishing? Appreciate it either way — thank you for the work you all put into TWiR every week!

nellshamrell:
Thanks for submitting this. We do not accept links solely to github repos or to crates.io pages. Please see our [guidelines for more info](https://github.com/rust-lang/this-week-in-rust/#projectstooling-updates). 

If you would do a write up of this project as a blog post or in the details of release notes, we would be happy to consider including it in a future issue.

ernestoariasdiaz:
Thanks for the feedback, that makes sense! I wrote up the project as a blog post: https://dev.to/ernesto_arias_148b35bc25d/-how-i-found-out-52-of-my-knowledge-graph-was-duplicates-and-what-i-did-about-it-3coh

It covers the Rust-specific engineering work (a recency bias bug in the semantic search, a gap-filter fix, and the engine migration to llama-server.exe + Gemma 4) along with the bigger lesson about catching a metric that only goes up but doesn't reflect real progress. Happy to have it considered for a future issue if it fits!

PR #8248, opened 2026-06-19
CI: Separate errors from warnings, post comments on either

### LINKED ISSUE #8154: Post comment when CI links/markdown checks fail to raise visibility of those failures
It would be nice for CI to also post a comment back to the PR when it fails (maybe just the first time it fails) to ask the author to take a look.

(Broken out of https://github.com/rust-lang/this-week-in-rust/issues/8148)

### LINKED ISSUE #8168: Separate errors from warnings in CI checks
So that the "soft warning" domain checks don't fail on main and conversely so our main CI will still fail on, e.g., invalid markdown.

jder:
This builds on #8153 to:
* Separate errors from warnings for the CI checks. The idea is that we can decide to merge things with warnings (i.e. it's an editorial decision) as opposed to errors which will fail CI.
* Everything that existed before #8153 are errors since they don't have false positives, except for the heuristic check on title truncation. The new checks for commonly-rejected links are warnings since they are sometimes ok.
* By default both scripts only show & fail on errors. (ie will pass on main)
* You can turn on warnings (and the related `--since` tag to filter down to recent ones) with `--show-warnings`. This still only fails for errors (arguably weird but this is because of next change).
* The CI check now posts a comment back to the PR if there are warnings or errors, and fails on errors. See examples of a [warnings only pr](https://github.com/jder/this-week-in-rust/pull/12) and [pr with errors](https://github.com/jder/this-week-in-rust/pull/13) for what this looks like.
* To make this easy, we make stdout be the content we'd post to the PR and use stderr for info/debug messages from the script (via normal python logging machinery) and move some existing prints to logging.
* Finally, we change the GitHub warning to look for root repo links, not any repo link.

Individual commits should be ~separately reviewable. 

Fixes #8154
Fixes #8168

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -11,6 +11,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: write
 
 jobs:
   Run-Markdown-Checks:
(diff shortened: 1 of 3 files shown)


nellshamrell (approved):
Looks excellent! Ty!

PR #8252, opened 2026-06-21
RustCurious lesson 9: Traits are Interfaces

benw:
This video explains traits in detail, covering associated types, trait inheritance, generic traits, and an explanation of the orphan rule involving a rabbit. We walk through Display, Debug, Iterator, IntoIterator, Mul, Index and IndexMut, ending with three practice exercises for learners to complete in the playground.


nellshamrell (approved):
Ty!

PR #8257, opened 2026-06-22
Rust Foundation Trusted Training & Rust Commercial Network Announcements

tinakrauss:
Added a link to the official program announcement under "Foundation" in the "Updates" section.

### DIFF
--- a/draft/2026-06-24-this-week-in-rust.md
+++ b/draft/2026-06-24-this-week-in-rust.md
@@ -40,6 +40,7 @@ and just ask the editors to select the category.
 ### Official
 
 ### Foundation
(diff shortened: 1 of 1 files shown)


tinakrauss:
I added a second announcement about the Rust Commercial Network to this PR. Thanks!

nellshamrell:
Hi @tinakrauss! I think there is a problem with the first link, this is what I see when I navigate to it

image

We're about to publish so I am closing this for now, but feel free to submit a new PR on the next issue!

tinakrauss:
@nellshamrell Thanks for the heads-up! The link is actually correct, but the site won't be live until the official program launch tomorrow morning. I thought the timing would work out since TWiR usually goes out later on Thursdays or Fridays, but I guess that was wishful thinking. No problem, though, I'll resubmit next week.

PR #8275, opened 2026-06-25
Add missing sentence for the crate of the week of entry 2026-06-24

sosthene-nitrokey:


llogiq (approved):
Thanks, good catch. I must've inadvertently removed the end of the sentence.

PR #8282, opened 2026-06-26
Add Axiom OS to Project/Tooling Updates

abhiprd2000:
Adding Axiom OS to Project/Tooling Updates. Independent no_std Rust kernel that does per-read file integrity by storing a per-4 KiB-block BLAKE3 leaf hash and verifying only the blocks a read touches. Per-read cost is proportional to bytes read instead of re-hashing the whole file. The repo includes the kernel and a benchmark harness. Happy to recategorize if the editors prefer a different section.

### DIFF
--- a/draft/2026-07-01-this-week-in-rust.md
+++ b/draft/2026-07-01-this-week-in-rust.md
@@ -45,6 +45,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


abhiprd2000:
Repo includes the kernel and benchmark harness, happy to adjust the title or section if preferred.

nellshamrell:
Thanks for submitting this. We do not accept links solely to github repos or to crates.io pages. Please see our [guidelines for more info](https://github.com/rust-lang/this-week-in-rust/#projectstooling-updates). 

If you would do a write up of this project as a blog post or in the details of release notes, we would be happy to consider including it in a future issue.

PR #8286, opened 2026-06-28
Add Rust and Allocator behaviour to Observations/Thoughts

sattva9:
How we diagnosed unexplained memory staying pinned near the top of our container limit in a Rust event stream processor — not a leak or fragmentation, and why the allocator matters.


nellshamrell (approved):
Ty!

PR #8279, opened 2026-06-26
Project/Tooling Updates: ds4-rs-metal (Rust + Metal DeepSeek-V4-Flash on Apple Silicon)

yijunyu:
Adds **ds4-rs-metal** under *Project/Tooling Updates*.

An open-source Rust + Metal inference engine for DeepSeek-V4-Flash on Apple Silicon that sustains ~300 tok/s prefill on an M1 Ultra 128GB Mac Studio — about +19% over the antirez/ds4 C reference, with a byte-identical +13.6% coming purely from MTLEvent-ordered command-buffer scheduling. Fully buildable with `cargo build`; a derivative of antirez/ds4 (MIT, vendored in-tree with attribution).

Repo: https://github.com/yijunyu/ds4-rs-metal

(Written by a human; honest framing — not fewer FLOPs, decode is a tie.)

PR #8288, opened 2026-06-29
Post PR comments from separate workflow with write permissions

jder:
The [previous code](https://github.com/rust-lang/this-week-in-rust/pull/8248) to post a comment back to contribution PRs when the checks failed worked only when the PR was opened from the same fork, because GitHub (very reasonably) prevents write access when running code from public forks. 

This code now uses a second workflow triggered by a [`workflow_run` event](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run) which reads the output from the test workflow and posts the comment. `workflow_run` workflows always use the default branch's content (not the PR branch) and so are trusted and get write tokens. (This case is one of the [examples given in the announcement of `workflow_run`](https://github.blog/news-insights/product-news/github-actions-improvements-for-fork-and-pull-request-workflows/)).

Here's [an example](https://github.com/jder/this-week-in-rust/pull/18) of this new code working in a cross-fork manner.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -11,7 +11,6 @@ on:
 
 permissions:
   contents: read
-  pull-requests: write
 
 jobs:
   Run-Markdown-Checks:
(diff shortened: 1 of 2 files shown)


jder:
@nellshamrell if you'd like to have a look. Would be great if whoever merges it could re-run tests on the PRs which failed [first one](https://github.com/rust-lang/this-week-in-rust/pull/8282) [second one](https://github.com/rust-lang/this-week-in-rust/pull/8279) to verify this now works.

nellshamrell (approved):
Thanks! Merging and will test on the two PRs you mentioned

PR #8289, opened 2026-06-29
Add post: a file format that destroys itself a little every time you open it

aravpanwar:
Adds a blog post under Observations/Thoughts.

The post is about decayfmt, a small Rust project of mine: a file format where every open permanently corrupts the file a little before it is displayed, with the decay rate read from a number in the filename and the corruption sourced from the OS CSPRNG so the sequence can never be reconstructed. The writeup is reflective rather than a Rust tutorial, so I placed it under Observations/Thoughts, but please move it if another category fits better.

Disclosure: this is my own project, and it is my single submission this week.

### DIFF
--- a/draft/2026-07-01-this-week-in-rust.md
+++ b/draft/2026-07-01-this-week-in-rust.md
@@ -47,6 +47,8 @@ and just ask the editors to select the category.
 
 ### Observations/Thoughts
 
+* [A file format that destroys itself a little every time you open it](https://www.aravpanwar.com/writing/a-file-format-that-destroys-itself/)
+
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hi @aravpanwar! This looks neat and it's cool that you wrote this in Rust, but there's not enough Rust-specific content in the post. Could you pull some example Rust code from the project and talk about why you chose what you did? Or what you learned about Rust in the process?

We are about to publish this week, so I'm going to close it, but feel free to re-submit this to next week's issue if you add some more Rust-specific content to the write up.

PR #8293, opened 2026-06-29
Add kache storage blog post to Miscellaneous

emmanuelm41:
Adds a Miscellaneous entry linking a kache blog post (placed in Miscellaneous to match the earlier "Content-addressed Rust builds (or, what kache actually caches)" deep-dive in the 2026-05-27 issue).

* [AI and worktrees are filling our disks: kache storage, measured](https://kunobi.ninja/blog/kache-storage-worktrees)

A technical deep-dive, with real measurements from a Firefox build (which mixes Rust and C/C++), on how kache — a content-addressed build cache — shares compiled artifacts across multiple git worktrees zero-copy via reflinks/copy-on-write: restoring 13.5 GB without copying a byte and saving 27.4 GB of duplicate data across a run. Covers methodology, an sccache comparison, implementation boundaries, and reproduction steps.

PR #8316, opened 2026-07-01
Add kache storage/worktrees post to Project/Tooling Updates

jleni:
Adds an Observations/Thoughts entry for our latest kache post:

* [AI and worktrees are filling our disks: kache storage, measured](https://kunobi.ninja/blog/kache-storage-worktrees)

The post measures how kache's reflink-based, content-addressed cache cuts duplicate storage across multiple Firefox worktrees (47.2 GB → 19.9 GB, 27.4 GB shared). This is placed in Observations/Thoughts rather than Project/Tooling Updates since it's a data/analysis writeup rather than a release update.

Disclaimer: the charts in the linked post were generated with AI assistance.

### DIFF
--- a/draft/2026-07-01-this-week-in-rust.md
+++ b/draft/2026-07-01-this-week-in-rust.md
@@ -45,9 +45,9 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


emmanuelm41:
We initially put it here, but we are not entirely sure which is the best category.  https://github.com/rust-lang/this-week-in-rust/pull/8293

PR #8333, opened 2026-07-03
Fix list item for multicalc in 2026-07-01 issue 658

kpreid:
The item had no bullet and was on the wrong side of the template comments.


tzilist:
Thanks!

PR #8339, opened 2026-07-06
Add CFP for Dofigen project - issue #278

taorepoara:
Here is a CFP for the Dofigen project targeting the issue https://github.com/lenra-io/dofigen/issues/278

let me know is there is any modification need since it's my first contribution to your news letter.


nellshamrell (approved):
Ty!

PR #8349, opened 2026-07-07
Add `Maintainer spotlight: Gen Li (@rami3l)` Inside Rust blog post

Kobzol:
https://blog.rust-lang.org/inside-rust/2026/07/07/maintainer-spotlight-gen-li-rami3l/

### DIFF
--- a/draft/2026-07-08-this-week-in-rust.md
+++ b/draft/2026-07-08-this-week-in-rust.md
@@ -39,6 +39,8 @@ and just ask the editors to select the category.
 
 ### Official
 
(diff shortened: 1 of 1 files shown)


ericseppanen:
Thanks for the link! For what it's worth, TWiR editors always include articles from the Rust blog + Inside Rust, so there's probably no need to open a PR to get these included.

Kobzol:
Right, I was told that before, but last two times the blog post didn't make it and only appeared there the next week 😅 So I was trying to be proactive. Ok, I won't send official blog posts anymore :)

PR #8359, opened 2026-07-08
2026-07-08: Investigating why RustCrypto is slow: Deep dive into SIMD instructions and hardware acceleration

skerkour:


tzilist:
Thank you!

PR #8383, opened 2026-07-13
Add links to Embedded Rust projects in the draft

theembeddedrustacean:
Links for µFerris hardware highlight and oxidize xiao community program


nellshamrell:
Hello there! The "Rust, Embedded Rust, and µFerris: An Open Source Learner Board Built Around Seeed Studio XIAO" feels more like a sales pitch and less like a Rust article, so unfortunately we will not include that one.

The other link "Oxidize XIAO — An Embedded Rust Community Program" is fine. I'm going to edit this contribution to just include that link.

Thank you for contributing to TWiR!

PR #8394, opened 2026-07-15
Add red-team writeup to Observations/Thoughts

akavlabs:
Adding a write-up under Observations/Thoughts: a four-pass red-team of an open-source Rust LLM-security proxy — the detection-engineering angle (evasion gaps, egress/streaming DLP, and keeping false positives at zero with a benign-corpus discipline). Not a crate release; it's an article with code.


tzilist (approved):
Thank you!

PR #8398, opened 2026-07-15
Add 'We compiled our TypeScript parser to WASM' to Observations/Thoughts

ivancernja:
We compiled our Rust TypeScript parser to WebAssembly with wasm-pack so it runs in the browser (rather than just in the CLI). The post covers gating OS-dependent code behind a native Cargo feature so it drops out of the WASM build, swapping disk reads for an in-memory file resolver, and embedding the SDK at compile time.


tzilist (approved):
Thank you!

PR #8414, opened 2026-07-19
Add ferrovec to Project/Tooling Updates

singhpratech:
Submitting a new project for the upcoming issue's **Project/Tooling Updates** section.

**[ferrovec](https://github.com/singhpratech/ferrovec)** is a dependency-light HNSW vector index written in Rust that compiles cleanly to `wasm32-unknown-unknown` for in-browser semantic search. `serde` + `postcard` are the only deps (~33 KB gzipped wasm), it uses a deterministic seeded PRNG (no `getrandom`) and SIMD128 distance kernels, and it also runs natively as a plain crate. Published on [crates.io](https://crates.io/crates/ferrovec).

Live demo (runs offline): https://singhpratech.github.io/ferrovec/demo.html

Happy to have the editors recategorize if another section fits better. Thanks!

### DIFF
--- a/draft/2026-07-22-this-week-in-rust.md
+++ b/draft/2026-07-22-this-week-in-rust.md
@@ -45,7 +45,7 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


opeolluwa (commented):
Thank you for submitting this! However, the descriptive text for the link is too long, could you limit it to around 10 words?

PR #8425, opened 2026-07-21
Nominate rag-gate for Crate of the Week

ajanm007:
Nominating `rag-gate` for Crate of the Week.

- Crates.io: https://crates.io/crates/rag-gate
- Repo: https://github.com/ajanm007/rag-gate

`rag-gate` is an inference-time confidence middleware that intercepts 
OpenAI-compatible LLM streams and gates generation into ANSWER, ABSTAIN, 
or ESCALATE using mean token logprob confidence — a signal already computed 
for free during inference. No post-hoc LLM judge, no retrieval-score 
threshold. Based on selective prediction research (IC3AI 2025).

Self-nomination.

### DIFF
--- a/draft/2026-07-22-this-week-in-rust.md
+++ b/draft/2026-07-22-this-week-in-rust.md
@@ -65,9 +65,11 @@ and just ask the editors to select the category.
 
 ## Crate of the Week
 
-<!-- COTW goes here -->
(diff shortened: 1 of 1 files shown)


bdillo:
hey @ajanm007 thanks for the contribution, but COTW nominations get picked from here: https://users.rust-lang.org/t/crate-of-the-week/2704, submit the crate there to get it in the running!

PR #8436, opened 2026-07-22
Add article: hash-chained audit log fork analysis (Ed25519 + TS/Rust verifier parity)

sandeepshekhar26:
A failure write-up from a Rust project: an Ed25519-signed, hash-chained JSONL audit log (Rust signer, TypeScript verifier with byte-for-byte parity tests) that develops duplicate-parent forks under concurrent writers — 13 in a 5,436-line live log from daily use. 

Covers:
- Raw-line hashing vs canonical re-serialization
- Cross-language canonicalization traps
- Why chain breaks can't be read as tampering until the writer serializes chain-head reads
- LLM-assistance is disclosed in the article per our guidelines

Runtime is MIT.


homersimpsons:
It looks like this article does not exist anymore, should it be removed from TWiR also?

sandeepshekhar26:
> It looks like this article does not exist anymore, should it be removed from TWiR also?

Will add it back, i think the link is broken, will fix it.

sandeepshekhar26:
> > It looks like this article does not exist anymore, should it be removed from TWiR also?
> 
> Will add it back, i think the link is broken, will fix it.

Its fixed now

PR #8437, opened 2026-07-22
Add Comhad to Project/Tooling Updates

Eoin-McMahon:
Hi! I'd like to submit **comhad**, a ranger-style terminal file browser for S3 (and S3-compatible endpoints), for this week's Project/Tooling Updates section. Think Cyberduck, but in the terminal.

comhad demo

> Storage backends sit behind a single `StorageProvider` async trait, so adding another service (SFTP, GCS, ...) is a matter of implementing that trait, the goal is for comhad to grow into a Cyberduck-style multi-backend tool rather than staying S3-only.

The linked v0.1.0 release write-up goes into a couple of the trickier Rust bits (the async trait backend abstraction, and a tmux/iTerm2/Sixel terminal-graphics-detection quirk I had to work around with `ratatui-image`).

Thanks for considering it!


bdillo (approved):
ty!

PR #8449, opened 2026-07-23
Add exiftool-rs 0.7.0 to Project/Tooling Updates

Le-Syl21:
Adds a one-line entry under Project/Tooling Updates for the exiftool-rs 0.7.0 release — a pure-Rust reimplementation of ExifTool (no Perl, no system libraries) with byte-for-byte name/value parity on the test corpus. 0.7.0 adds localized PrintConv values under `-lang`.

Repo: https://github.com/Le-Syl21/exiftool-rs

### DIFF
--- a/draft/2026-07-29-this-week-in-rust.md
+++ b/draft/2026-07-29-this-week-in-rust.md
@@ -45,6 +45,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


Le-Syl21:
Thanks — good catch by the bot. I've repointed the link from the bare repo to the 0.7.0 release write-up, which goes beyond the changelog: it explains why a flat translation table over-translates ExifTool's PrintConv values, the group-scoped override design that fixes it, and the XMP edge case, with the honest scope caveats.

For transparency per your LLM-content guidelines: the release notes were written with LLM assistance (noted on the release page itself).

PR #8452, opened 2026-07-23
Fixing incorrect Australian state/territory codes in meetup automation script

### LINKED ISSUE #8208: Australian state/territory codes being mangled again
I fixed this in #8014 in April, but it’s started again in the latest TWiR (“Brisbane City, QL, AU”). I won’t offer a patch this time, because there’s clearly a problem that needs to be addressed somewhere else, and patching symptoms alone is useless.

KannanPalani57:
### Description
Australia's state abbreviations aren't a fixed length — they're a mix of 2 and 3 letters depending on the state.  
Good examples of each:
Two-letter codes: WA (Western Australia), SA (South Australia), NT (Northern Territory)
Three-letter codes: NSW (New South Wales), VIC (Victoria), QLD (Queensland), TAS (Tasmania), ACT (Australian Capital Territory)

This PR fixes the issue by using a fixed city/state lookup map, where each city is mapped explicitly to its correct official abbreviation

Closes #8208

### DIFF
--- a/tools/events/meetup-automation/city_to_state_map.py
+++ b/tools/events/meetup-automation/city_to_state_map.py
@@ -1,0 +1,57 @@
+# Mapping of known cities to their state codes
+CITY_STATE_MAP = {
+    # New South Wales
+    "Sydney": "NSW",
+    "Newcastle": "NSW",
+    "Wollongong": "NSW",
+    "Central Coast": "NSW",
+
+    # Victoria
+    "Melbourne": "VIC",
+    "Geelong": "VIC",
+    "Ballarat": "VIC",
+    "Bendigo": "VIC",
+
+    # Queensland
+    "Brisbane": "QLD",
+    "Gold Coast": "QLD",
+    "Sunshine Coast": "QLD",
+    "Townsville": "QLD",
+    "Cairns": "QLD",
+
+    # Western Australia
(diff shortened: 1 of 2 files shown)


tools/events/meetup-automation/city_to_state_map.py
```diff
@@ -0,0 +1,57 @@
+# Mapping of known cities to their state codes
+CITY_STATE_MAP = {
+    # New South Wales
+    "Sydney": "NSW",
+    "Newcastle": "NSW",
+    "Wollongong": "NSW",
+    "Central Coast": "NSW",
+
```
bdillo, line 8: Normally we only write the state codes for US events (don't think there's any reason in particular for that, but that's just the current format), so generally these just get left out anyways.

If we did want to do this though, why hardcode a list of city names? Shouldn't we just take a set of state names and expand them out if needed, rather than having a static list of state names? Like if there's a city outside this list, the proper state code wouldn't be used, right?

KannanPalani57: Make sense, From  meetups API data, we don't get the state/territory code for the Australia country.  And hardcoding the state codes is not the best approach.

KannanPalani57:
Since hardcoding state code is not the feasible solution. I am closing this pr.

PR #8462, opened 2026-07-24
Add Schobio to Project/Tooling Updates

Kamayuq:
Adds Schobio to Project/Tooling Updates.

Schobio is a thin Rust frontend over the Schobi C work-stealing scheduler,
written as a proof of concept for what Rust's type system can enforce on top of
a C runtime: borrowed closures instead of `Arc`, a `catch_unwind` firewall at
every FFI entry point, and a `GroupBarrier` that is `!Send + !Sync` so a wait
from outside its fiber group fails to compile rather than segfaulting in C.

The C core is the real engine — an eager lenient fork/join scheduler that clears 455 Mtasks/s
at 0.33 µs p50 on an M4 Max (~10× oneTBB), exhaustively verified under CBMC,
GenMC and Relacy — with Kani and loom proving the Rust glue on top. The README
is a walkthrough rather than a changelog: the four C contracts the `unsafe`
blocks depend on, the proof ladder, the debuggability story (frame-pointer-clean
fiber unwinds, a `.natvis` view of parked tasks, profiler markers), and
benchmark methodology against rayon, chili, fork_union and bevy_tasks including
the events where Schobio loses and why.

### DIFF
--- a/draft/2026-07-29-this-week-in-rust.md
+++ b/draft/2026-07-29-this-week-in-rust.md
@@ -45,6 +45,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hello there! Thanks for submitting this. There are a couple of things we need before we can include it:

1) I'm getting a 504 Gateway error when I try to access that link
2) The descriptive text for the link is too long, could you limit it to around 5 words?

nellshamrell:

PR #8487, opened 2026-07-29
Add an article about Fluent

mrinalwadhwa:
Added an article about Fluent, an open-source, self-improving software factory built in Rust

https://mrinal.com/articles/how-i-built-a-self-improving-software-factory/
https://github.com/mrinalwadhwa/fluent

### DIFF
--- a/draft/2026-07-29-this-week-in-rust.md
+++ b/draft/2026-07-29-this-week-in-rust.md
@@ -47,6 +47,8 @@ and just ask the editors to select the category.
 
 ### Observations/Thoughts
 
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hello there! Unfortunately, there is not enough Rust specific content for this to be included in This Week in Rust. Check out our [guidelines](https://github.com/rust-lang/this-week-in-rust#what-do-we-look-for-when-considering-whether-to-include-something-in-this-week-in-rust) for more information about what we look for in articles that we include.

mrinalwadhwa:
Hey @nellshamrell thank you for the feedback. Is there a category for things built in Rust even if the article isn't Rust centric.

Totally understand if it's not a fit.

PR #8502, opened 2026-07-30
Add Oxi GDI rounding walkthrough

Ryujiyasu:
Self-submission. A walkthrough of reproducing Microsoft Word's GDI integer-pixel quantization in Rust — component-wise rounding order, context-dependent rounding direction, and where a measured coefficient replaces a formula. Includes the Word COM / Win32 GDI / PDF oracle setup used to derive the rules.


Ryujiyasu:
I wrote this in Japanese first and translated it myself; happy to add a note if you'd prefer disclosure of any tooling used.

PR #8509, opened 2026-07-31
Add hooklog on Topcoat walkthrough

JuanMarchetto:
A build log of hooklog, a self-hosted webhook inspector, built on Topcoat six days after that framework went public. It records the commands, the compiler and runtime errors, and the measured build times, request counts and page latencies as they happened.


KannanPalani57 (approved):
Thank you!

PR #8522, opened 2026-08-03
Add CFP for Dofigen project - issue #481

taorepoara:
Here is a CFP for the Dofigen project targeting the issue https://github.com/lenra-io/dofigen/issues/481


KannanPalani57 (approved):
Thank you!

PR #8532, opened 2026-08-04
Add RustConf ticket pricing update

tinakrauss:
Add RustConf ticket pricing update under "Official" - "Foundation"

### DIFF
--- a/draft/2026-08-05-this-week-in-rust.md
+++ b/draft/2026-08-05-this-week-in-rust.md
@@ -40,7 +40,8 @@ and just ask the editors to select the category.
 ### Official
 
 ### Foundation
-
+- [RustConf 2026](https://rustconf.com): Employer-paid ticket pricing increases on August 11 — [register now!](https://rustconf.com/register/)
+  
 ### Newsletters
 
 ### Project/Tooling Updates
--- a/draft/2026-08-05-this-week-in-rust.md
+++ b/draft/2026-08-05-this-week-in-rust.md
@@ -40,7 +40,7 @@ and just ask the editors to select the category.
 ### Official
 
 ### Foundation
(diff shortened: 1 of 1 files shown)


tinakrauss:
The links I added are accurate. Can they please be accepted as is? If not, please let me know what needs to be changed.

ericseppanen:
Including the same link in more than one issue will break our CI, so accepting the links as-is will cause persistent failures in our tooling. Unfortunately, right now we don't have an escape hatch to allow duplicate links.

My recommendation would be to link to a blog post announcing the pricing update, if one exists. Also note that our preferred style is one link per line, i.e. `[some text](a description)`.

Also, this should probably not go in the Foundation section. As far as I'm aware we don't treat RustConf messaging as part of the Foundation. @nellshamrell does this sound right?

tinakrauss:
@ericseppanen Thank you for the explanation! I tried to remove the /register link to avoid the duplicate, and was only keeping the main URL (https://rustconf.com/), which I don't believe has been used before. However, I'm getting this error:

Sorry, this diff is unavailable.
The repository may be missing relevant data. Please [contact support](https://github.com/contact) for more information.

Any advice on how to fix this?

Also, please feel free to move this to any other section if the Foundation section is not the right place. Thank you!

ericseppanen:
I don't think this submission really fits in the "Community Links" section. I understand you want to share the word about conference pricing changes, but that's not really what the links section is for.

There is an events section, and RustConf is already included there. Unless @nellshamrell or the Events editors want to weigh in, I don't think this type of messaging fits our current format.

nellshamrell:
Hi @tinakrauss! Apologies for the delays in my response! This isn't the kind of thing we feature in the Community Links section (RustConf is already present in the events section). If you were to create a Foundation blog post highlighting the price change, that is something we would consider accepting.

Thank you for all you do for RustConf and the Rust Community!

PR #8533, opened 2026-08-04
Add Kuala Lumpur Rust Meetup to events list

pickfire:


pickfire:
By the way, I am wondering, how come we always missed this? Usually we create events in the rust community calendar 2-4 weeks ahead of time, does the script not re-processed the old events? When does the script process the events?

If there is anything that I can help to do on my end to make sure the events always get posted instead of manually submitting PR to add the event in, please let me know ya. Thanks.

KannanPalani57:
This event included in the bulk events update pr #8529 .

KannanPalani57:
Events marked in the calender will be updated in the newsletter by the events editor manually by adding it.  We use the script to retrieve the rust events from meetup through their API's.  This is the code for the [meetup_script](https://github.com/rust-lang/this-week-in-rust/blob/main/tools/events/meetup-automation/main.py).

So , events that are from meetup.com automaticalled fetched by our script.

KannanPalani57:
Thank you!

pickfire:
So only events from meetup.com? Rust Malaysia we don't use meetup.com previously because I have issues logging into meetup.com and we need to pay for it (we are non-profit group and doesn't have sponsors lately, I rather pay for food than meetup), so we only update rust calendar only, does it take from other sources like eventbrite or google calendar?

PR #8537, opened 2026-08-05
Improve submerge error handling

jder:
* When a PR targets the wrong draft, treat as a non-fatal error (we could later just handle this by extracting the added link if it is common)
* When a fatal error occurs, allow bypassing that PR with --keep-going.

There is then a bit of code movement + cleanup in the tests due to moving the file-checking location.


nellshamrell (approved):
Ty so much!

PR #8538, opened 2026-08-05
Add FamilyClaw v1.3.0 to Project/Tooling Updates

Sisuthros:
Adding FamilyClaw v1.3.0 to this week's Project/Tooling Updates.

FamilyClaw is a crash-safe Rust runtime for AI agents that perform real
external actions — charging cards, sending email, deleting objects. The core
claim is at-most-once external dispatch across tested crash and replay
windows, verified by a 60-second reproducible proof (`scripts/crash-proof.sh`,
no API keys, no network) that SIGKILLs the runtime after a side effect fires
but before the durable commit record is written, then checks that the
side-effect count stays at 1. Running the same harness against the pre-fix
code path double-fires, so the proof measures behaviour rather than printing a
constant.

The linked release notes are a full write-up — highlights, security fixes, an
explicit "known limitations" list, and the local verification commands the
release was cut from — rather than a bare changelog, per the section
guidelines. The repository also contains a crash-safety benchmark against
LangGraph at its strongest durability configuration
(`bench-competitors/langgraph/`) with its own methodology and honesty caveats,
for readers who want the raw comparison.

MIT licensed, `unsafe` forbidden across the workspace, prebuilt Linux and
Windows binaries attached to the release.

Disclosure: I am the project author, and this is my one project submission for
the week per the contributor limit.

Disclosure: the release notes and much of the project documentation were
written with substantial AI assistance, as were parts of the codebase. Flagging
this explicitly per the repository's guidance on LLM-written content. The
benchmark numbers, the crash proof and the test results are reproducible from
the repository by anyone who runs the commands.

### DIFF
--- a/draft/2026-08-05-this-week-in-rust.md
+++ b/draft/2026-08-05-this-week-in-rust.md
@@ -45,6 +45,8 @@ and just ask the editors to select the category.
 
 ### Project/Tooling Updates
 
+* [FamilyClaw v1.3.0: at-most-once external side effects for AI agents, proven by a 60-second crash test](https://github.com/Sisuthros/familyclaw-oss/releases/tag/v1.3.0)
+
 ### Observations/Thoughts
 
 ### Rust Walkthroughs


KannanPalani57:
Thank you for submitting this! However, the descriptive text for the link is too long, could you limit it to around 5 words?

KannanPalani57:
We're about to publish shortly! Feel free to submit a follow-up PR for next week's newsletter

PR #8557, opened 2026-08-08
Rust Melbourne: use Luma instead of Meetup

metasal1:
## Summary
Rust Melbourne has moved event registration from Meetup to Luma.

- Draft + latest issue Oceania entries now link to:
  - Calendar: https://luma.com/rustmelbourne
  - August meetup: https://luma.com/d0rndgyv
- Removed `https://www.meetup.com/rust-melbourne/events/` from `tools/events/rust-meetups.json` so the Meetup automation does not keep re-injecting Meetup URLs (that tooling only supports `meetup.com` hostnames).

## Note
Historical issues are left unchanged. Future Melbourne events can be added in the draft the same way other Luma groups are listed (e.g. Rust Girona / Rust SP).

## Test plan
- [x] Confirm Luma calendar and August event URLs resolve
- [x] `rust-meetups.json` still valid JSON
- [ ] TWiR editor review

### DIFF
--- a/content/2026-08-05-this-week-in-rust.md
+++ b/content/2026-08-05-this-week-in-rust.md
@@ -383,8 +383,8 @@ Rusty Events between 2026-08-05 - 2026-09-02 🦀
     * [**Rust-Atl**](https://www.meetup.com/rust-atl/events/313539331/)
 
 ### Oceania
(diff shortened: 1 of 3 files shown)


KannanPalani57 (approved):
Thank you for making the changes too in the rust meetups json file!

PR #8579, opened 2026-08-11
Add "Workshop: Add tests to an open source Rust project"

szabgab:
It was postponed


KannanPalani57:
Do we need to include this event?

szabgab:
Well, this is the new date of the event, but with the same link as before.

KannanPalani57 (approved):
Thank you!

Issue #8575, opened 2026-08-10
Changes to the Projects/Tooling Section


nellshamrell:
Hello all,

When I became the editor of This Week in Rust in early 2020 we received around 7-10 link submissions via pull requests for each issue. Now, 6 years later, we receive 40+ pull requests per week with link submissions. The majority of these submissions are for the Projects/Tooling Updates section. There are two large reasons for this. First, Rust usage has grown massively (which is great!) and more and more people are creating projects that do very cool things with Rust. Second, the widening use of LLMs for writing Rust code has significantly lowered the barrier (both in terms of time as well as effort) for generating Rust projects. 

We don't take a position on whether you should or should not use LLMs to write code or other content. There are many cases where they are genuinely useful - such as translation or summarizing the commits that go into a release. However, the number of submissions for this section each week (all of which are manually reviewed) has become unsustainable for a group of volunteer editors. 

We need the help of the community in reviewing and surfacing interesting projects for inclusion in This Week in Rust. We considered a few different options and ultimately settled on **pulling links for the Projects/Tooling Updates section only from [r/rust](https://www.reddit.com/r/rust/)**. There is already a very active community there and we need their help.

What does this mean? **Starting after this week's issue (Issue 664), we will no longer be accepting links for Project/Tooling via pull request (other sections of the newsletter will remain unchanged). If you want a Project/Tooling update to be considered for This Week in Rust, you must post it to r/rust (and you must follow their community guidelines)**.  We will be relying on the community there to upvote the Projects/Tooling updates that they find the most interesting. Whether or not a link is included in This Week in Rust will remain at the discretion of the This Week in Rust editors, but we will certainly consider community input in our process.

We know this is a change to a process that has existed for many years. However, it is absolutely necessary to prevent overwork and burnout of the This Week in Rust editors. We appreciate your understanding!

Shnatsel:
Do I understand correctly that once the link is submitted to r/rust, no further action is needed for it to be considered for inclusion in TWIR?

nellshamrell:
@Shnatsel - correct, though it is not a guarantee that it will be included. Our editors monitor r/rust and will select links for the Projects/Tooling section from there only. You will (of course) need to follow r/rust's community guidelines for what is allowed to be posted there.

PR #8590, opened 2026-08-12
Add renew 0.1.1 to Project/Tooling Updates

CagdasErturk:
Adds one line to Project/Tooling Updates in the current draft, linking the first release of [renew](https://github.com/renew-engine/renew).

renew is a game engine in Rust whose simulation core is bit-deterministic by construction: a fixed timestep over integer nanoseconds, arithmetic in Q47.16 fixed point rather than floating point, no wall-clock reads, no unseeded randomness, and component storage whose iteration order is defined by slot. Each platform emits its simulation digests in CI and a final job compares them against one another, so the cross-platform claim is discharged by targets agreeing with each other rather than against a committed constant.

0.1.1 is the first release: 27 crates on crates.io under the `renew-` prefix, Apache-2.0, documentation on docs.rs. Five crates are core; the rest are optional and removable, and CI builds and tests one configuration per removed crate.

It is in early development, with no module yet at `stable` maturity, and the linked release says so plainly rather than overselling.

This is a project submission rather than an article, and it is my only project submission this week.

### DIFF
--- a/draft/2026-08-12-this-week-in-rust.md
+++ b/draft/2026-08-12-this-week-in-rust.md
@@ -47,6 +47,7 @@ and just ask the editors to select the category.
 
 * [GRIT 1.1: type-check your quantized tensors](https://singhpratech.github.io/grit-datatype/)
 * [git-cache-proxy: read-only cache for git](https://rolandsdev.blog/posts/caching-git-clones-across-a-slow-network/)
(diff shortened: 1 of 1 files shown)


nellshamrell:
Hello there!

Please know that we are making changes to the link submission/review process for the Projects/Tooling section starting with next week's issue. Please see [this GitHub issue](https://github.com/rust-lang/this-week-in-rust/issues/8575) for more information.

PR #8601, opened 2026-08-13
Add axioma-core to project updates

axmalgorithm-ops:
axioma-core v0.1.0 — no_std, zero-allocation compression engine

Crate: https://crates.io/crates/axioma-core
Repository: https://github.com/axmalgorithm-ops/axioma
Author: Murad Abbasov

axioma-core is a small compression engine built for embedded and resource-limited targets. It runs with #![no_std], allocates nothing during encoding or decoding, and avoids division operations entirely, which keeps it usable on microcontrollers without hardware dividers.

The pipeline is intentionally straightforward: a delta preprocessor for numeric sequences, a bounded LZ77 sliding-window matcher, an order-1 adaptive context model, and a 32-bit binary range coder. No unsafe code. Decoding never panics — corrupted input returns an error.

On structured telemetry and log data it typically achieves around 3.48x compression while keeping the full encoder state under 70 KiB. It fits scenarios like firmware log storage, LoRa/UART sensor streams, or any place where gzip/zlib are too heavy.

If it looks useful, it could be a good candidate for Crate of the Week.

### DIFF
--- a/draft/2026-08-19-this-week-in-rust.md
+++ b/draft/2026-08-19-this-week-in-rust.md
@@ -44,6 +44,7 @@ and just ask the editors to select the category.
 ### Newsletters
 
 ### Project/Tooling Updates
+* [axioma-core](https://crates.io/crates/axioma-core) — no_std, zero-allocation compression engine for embedded targets
 
 ### Observations/Thoughts
 


nellshamrell:
Thanks for submitting this link! However, we are no longer accepting pull request submissions for the Project/Tooling Updates section (see [here](https://github.com/rust-lang/this-week-in-rust/issues/8575) for more details). Our editors monitor [r/rust](https://www.reddit.com/r/rust/) and will consider links that are posted there. Please make sure to follow r/rust's community rules when you are there!

PR #8628, opened 2026-08-18
Add CFT, FCP, MPC, and RFC for TWiR 665

tzilist:
This PR also adds an experimental generator script to do the work we used to do by hand.


nellshamrell (approved):
Ty!

PR #8631, opened 2026-08-19
Add MCP browser bot-detection study to Observations/Thoughts

pitiflautico:
Adds a reproducible study comparing two MCP browser tools (NeoBrowser, a Rust MCP server driving real Chrome via CDP, vs Playwright MCP) against live public bot-detection sites, with raw data and methodology published in the linked repo.

Entry:
* [I ran two MCP browser tools against live bot detection. Here's the honest table.](https://pitiflautico.github.io/neobrowser/study.html)

### DIFF
--- a/draft/2026-08-19-this-week-in-rust.md
+++ b/draft/2026-08-19-this-week-in-rust.md
@@ -47,6 +47,8 @@ and just ask the editors to select the category.
 
 ### Observations/Thoughts
 
+* [I ran two MCP browser tools against live bot detection. Here's the honest table.](https://pitiflautico.github.io/neobrowser/study.html)
+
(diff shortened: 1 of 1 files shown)


alex-semenyuk (approved):

opeolluwa (commented):
Hi @pitiflautico,

Thank you for writing to TWiR.
Unfortunately, the article doesn't provide enough context for us to consider it. I'm going to have to close it.

Please take a look at [What do we look for when considering whether to include something in This Week in Rust](https://github.com/rust-lang/this-week-in-rust/blob/main/README.md#what-do-we-look-for-when-considering-whether-to-include-something-in-this-week-in-rust)

Happy to receive your entry in the future.

PR #8641, opened 2026-08-20
Use PR author write permission rather than association

jder:
This seems like a more precise check for whether we should allow adding of these links given the [failure here](https://github.com/rust-lang/this-week-in-rust/actions/runs/32330055881/job/96308830743?pr=8639).


nellshamrell:
Thank you!

PR #8650, opened 2026-08-24
Add Rust Glancer, a new Rust LSP

popzxc:
This was previously discussed on r/rust and hackernews, so I hope it's appropriate to add the announcement here as well.


popzxc:
Ah ok, makes senes (sad that it became necessary, but I 100% understand).
It was already posted to r/rust, so I guess all good.

PR #8661, opened 2026-08-26
add crux to Project/Tooling Updates

Sal-ami:
Adds crux, a CLI that finds the commit which changed a program's behavior and minimizes the flip commit to the lines that caused it. Written in Rust, 2 MB binary, published on crates.io as crux-finder.


Sal-ami:
Understood, closing. Per #8575 I'll post the project to r/rust instead for the editors to pick up.

PR #8690, opened 2026-09-01
Add GPUI article and Rust variables and types video link

MrSheerluck:


tzilist:
@MrSheerluck thank you for submitting. Generally, we only allow one link per submission per week. On that note, I will submit your article, but, feel free to submit again next week with your video. Thank you!

PR #249, opened 2016-07-07
Updates for twir-138

matthiasbeyer:
Feel free to cherry-pick any commit you want to.

Also, if I did something wrong, feel free to yell at me! :smile:

PR #8694, opened 2026-09-01
CFP: sysknife

vladimirrott:
Three tasks from sysknife, an MIT-licensed MCP daemon that runs Linux administration actions behind a typed catalogue and a signed audit chain.

* `easy` (#234): a ufw action takes a rule number that no read action in the catalogue can produce, so the planner is pointed at a value it cannot obtain.
* `medium` (#250): the PID-reuse check on the D-Bus caller collapses two different `getsockopt` failures into one, so a peer that exited during the `/proc` read is treated the same as a kernel that never supported the pin.
* `hard` (#249): the signed trail records which account asked for a change and never which one approved it, across two storage backends.

All three are open, unassigned, carry a difficulty section, and link CONTRIBUTING.md. Public tracker, no CLA.


nellshamrell (approved):
Thank you!

Issue #668, opened 2018-07-04
In the News & Blog Posts section would it be possible to get links to any discussion (if there is any) on reddit, hackernews etc? 


hgallagher1993:

nasa42:
Starting next week, we plan to add reddit links as an experiment and watch for community's feedback. Thanks for the idea!

(PS: Also [discussed here](https://www.reddit.com/r/rust/comments/97czdf/this_week_in_rust_247/e47977d/)).

PR #8704, opened 2026-09-03
Add SyntaxCue audio-capture walkthrough to Rust Walkthroughs

ZetBaur:
Adding a technical walkthrough to the Rust Walkthroughs section of the 2026-09-09 draft (#668).

The article covers the full local-audio pipeline of a Tauri/Rust desktop app: macOS CoreAudio Process Tap (via a Swift sidecar, since the API has no Rust binding), Windows WASAPI loopback reached directly from Rust via COM, voice-activity detection, and local Whisper transcription — with the actual APIs used and measured latency numbers.

Link: https://syntaxcue.com/how-syntaxcue-captures-system-audio/

Disclosure: I'm the author of the article and the app it describes (SyntaxCue, an AI interview-prep copilot). Happy to have an editor move/recategorize the entry if it fits better elsewhere, or drop it if it's not a fit.

### DIFF
--- a/draft/2026-09-09-this-week-in-rust.md
+++ b/draft/2026-09-09-this-week-in-rust.md
@@ -52,6 +52,8 @@ and just ask the editors to select the category.
 
 ### Rust Walkthroughs
 
+* [How SyntaxCue Captures System Audio Locally](https://syntaxcue.com/how-syntaxcue-captures-system-audio/)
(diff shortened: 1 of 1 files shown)


ericseppanen:
There is not enough Rust content in this article to justify a spot in TWiR.

PR #8706, opened 2026-09-03
Add 'The undo has to exist before the write does' to Observations/Thoughts

mahirhir:
One line into Observations/Thoughts for the 2026-09-09 draft.

https://dev.to/mahirhir/the-undo-has-to-exist-before-the-write-does-46on

It is about an ordering invariant in a Rust engine I work on: the inverse of a write is constructed and escrowed before the write is applied, rather than reconstructed afterwards from a diff. The post covers why the after-the-fact version cannot be made sound when a crash lands between apply and record, and what the engine gives up to get the stronger ordering.

Self-submitted and my own writing. It is a design post rather than a release announcement; the crate is alpha and the post says so. Happy to move it to another section or drop it if Observations/Thoughts is not the right home.


alex-semenyuk (approved):

ericseppanen:

PR #8715, opened 2026-09-08
Blog post: A subscribe button does not need a Worker

EmilLindfors:
https://lindfors.no/blog/newsletter-on-my-own-server/

The newsletter moved off Cloudflare Workers and onto the mail server in a day: one Rust service, a Postgres with four tables, addresses sealed at rest, and a record of which issue went to whom. Why a mail store as a database was the right workaround in February and the wrong one now, what the move cut from the attack surface, and what the research on LLM-written code says about doing this with a coding agent without the complexity creeping back.

Author submission; feel free to move it to another section.


ericseppanen:

PR #8734, opened 2026-09-09
add a warning to the draft template regarding Projects/Tooling PRs

ericseppanen:
We continue to get PRs for this section, so this will hopefully catch a few of those before they are submitted.


bdillo (approved):
good idea!

PR #8738, opened 2026-09-10
Added link to version 3.0.0 of Rust Projects - Write a Redis Clone

lgiordani:
Hi!

I added a link to the new version of my free book "Rust Projects - Write a Redis Clone"
The book is free both in HTML and PDF forms. Since downloading the PDF version requires an email I added here only the HTML version, that is freely accessible.

Thanks!


lgiordani:
Hi! I added a tag `#3.0.0` to the URL to differentiate it from the previously added one. Not sure if you are happy with this, as it basically circumnavigates the check, but the content of the link is actually a new version of the book.

Let me know! Thanks!

PR #8750, opened 2026-09-15
Add link to awesome rust papers

Evian-Zhang:
This is a list of academic papers about Rust. The list is built by human, and all LLM-involved parts are verified carefully by human. I don't know if it is appropriate to put it in the miscellaneous category. It seems to not fit in the research category, since itself is not a paper.


extrawurst:
This submission points at an aggregator of articles that *might* individually qualify for an entry but a platform itself is not. it would be like adding a link to `/r/rust` on reddit in the misc section. feel free to submit a rust article for inclusion you deem qualified for entry

PR #8752, opened 2026-09-15
Add 'Teaching git diff to read JSON and YAML' to Rust Walkthroughs

dimanovikov:
A walkthrough of wiring a Rust CLI into git as an external diff driver and a textconv filter: the seven arguments git passes, why a non-zero exit aborts the whole diff, detecting the format from the real path instead of temp files, and why the fallback reads bytes (`std::fs::read`) rather than `read_to_string`.

Written by me. If Walkthroughs isn't the right section, please move it wherever it fits.


ericseppanen:

PR #8761, opened 2026-09-18
Add ENTITY Protocol Rust conformance CFP

blackmore-technology-group:
Adds a Call for Participation item for an independently authored Rust implementation of ENTITY Protocol 1.0.

The linked task:
- is open source under Apache-2.0;
- has a public issue tracker;
- is explicitly labeled hard;
- links contribution guidance;
- requires an independent Rust implementation in its own repository;
- targets a sealed clean-room protocol conformance kit rather than the BTG Python reference implementation.

This is a contributor recruitment item, not a project/tooling update.


ericseppanen:
This reads like a set of directions for an agent. If you're searching for open source collaborators to contribute their valuable time, this message doesn't seem very welcoming.

blackmore-technology-group:
Thanks for the feedback. You were right that the original issue read too much like a specification and not enough like an invitation to a contributor. I have rewritten it to make the contributor's freedom and the collaborative nature of the work clear, while keeping the detailed conformance requirements separate for anyone who decides to take it on.

blackmore-technology-group:
One follow-up: I simplified the linked Rust issue further so the contributor invitation is now the entire opening section, almost exactly in the form suggested here. The detailed conformance checklist and resources are now secondary material under a collapsible section, rather than the voice a potential contributor sees first. I also corrected the formatting/encoding in that issue.

nellshamrell:
Hello there! Thank you for submitting this. This task is bigger than what we are looking for in this section - we're focused more on well scoped issues that can be picked up by a contributor quickly. This feels more like a very large project. I recommend taking this to r/rust on reddit.

blackmore-technology-group:
Thanks — that makes sense. I’ve taken the feedback on scope and broken the work into much smaller, bounded contributor tasks in the project rather than trying to fit the full implementation challenge into this section. I’ll keep this-week-in-rust focused on work that can be picked up quickly and take the broader project outreach elsewhere. Appreciate the direction.

PR #8764, opened 2026-09-19
Add 'Why datadiff matches arrays by key instead of computing tree edit distance' to Rust Walkthroughs

dimanovikov:
A walkthrough of the Rust side of a structural diff tool: a BTreeMap-based value tree that makes key order irrelevant, matching array elements by key with let-else and ? falling back to index matching, comparing Int/UInt/Float numbers, and three CSV parsing pitfalls (`"00544".parse::<i64>()`, `f64::from_str` accepting `nan`/`inf` case-insensitively, integers past i64 losing digits as f64), plus a test race caused by SystemTime's microsecond resolution on macOS.

Follow-up to #8752, which was about git integration rather than Rust. If Walkthroughs isn't the right section, please move it.


opeolluwa (approved):
@dimanovikov, would you

